PR #4, opened 2012-03-29
Module deps

petebacondarwin:
Removed requires param from angular.module() calls to prevent overwriting previously defined modules.


edwardhotchkiss:
Awesome Pete :P

PR #16, opened 2012-05-18
Fixed uiEvent isString check

bignall:
It said angular.isString(event) but should be angular.isString(events)


bignall:
Even with this correction, there's still a problem.  Using a uiEvent gives the following message:

Uncaught TypeError: Object #<Object> has no method 'apply'

This happens on your example on the angular-ui home page as well.

PR #29, opened 2012-06-28
updated testing to 1.0

dandoyon:
fixed ui-date to use 1.0
removed coffee code for currency and template
moved test lib for angular under a version directory (angular-1.0.1), I think that this will make things
easier as new releases come out. Not sure.
NOTE: uiMask may not be working, I accidentally checked in without confirming it works on 1.0
I have not yet tried using it will check it out tomorrow.


dandoyon:
I still have work to do on the template stuff, so don't trust as complete

PR #35, opened 2012-07-07
Keypress mod

borodiychuk:
Keypress mod. Keys combinations support and alternative syntax.


borodiychuk:
When I pushed some fixes, should I close this pull request and start new one?

ProLoser:
No it's fine. As Peter (I think) pointed out, all your new commits are being appended to this pull request.

Speaking of which, do you mind switching to use $observe? I don't think there's a reason not to and the way it's coded now causes a lot of unnecessary overhead processing.

borodiychuk:
I take small time-out, need to learn about $observe.

borodiychuk:
> Speaking of which, do you mind switching to use $observe? I don't think there's a reason not to and the way it's coded now causes a lot of unnecessary overhead processing.

Ok, so as for $observe. It would make sense if our attribute `ngKeypress` contained any interpolated thing like `{{var}}`. But our attribute (either object or string) never gets changed. So, if we switch to `attr.$observe`, we will never get our funtion executed, because attribute's text is always constant.

ProLoser:
The documentation isn't consistent. It says it's used for interpreting {{}} however I could have sworn I use it elsewhere for non-encapsulated values. In addition, the example code they provide has ngModel, which does **not** use {{}}. Need to get the core team to clarify this.

Pinging @petebacondarwin because his 2 directives (mask and date) use $observe and I don't believe they leverage `{{}}` at all.

I recommend testing this first. Although the code is not likely to change, the value may very well change. If nothing changed at all then there would be no reason to evaluate on every single keypress, which is the main thing I'm hoping to avoid.

borodiychuk:
> If nothing changed at all then there would be no reason to evaluate on every single keypress, which is the main thing I'm hoping to avoid.

But that is wrong. The handler should be evaluated on each keypress to find out if the pressed key matches any defined key combination (but we will bother $eval only when we found matched key combination).

And as I told above, not all keys change the value of input. Enter and Escape don't for example, but the most common case is to clear input on escape and process its value on enter.

modules/directives/keypress/src/keypress.js
```diff
@@ -1,19 +1,74 @@
...
+				39: 'right',
+				40: 'down',
+				45: 'insert',
+				46: 'delete'
+			};
+
+			elm.bind('keydown', function(event) {
+				var params, paramsParsed, expression, keys;
+				try {
+					params = scope.$eval(attrs.uiKeypress);
```
ProLoser: I'm saying this line is being executed every single time the user hits a key. You are effectively evaluating every single time someone clicks any key on the keyboard for this element. Is there any way this can be moved outside of the keyup binding like it used to be?

_You_ are referring to [line #69](#L0R69) which is an additional evaluation.

borodiychuk: Oh, right. I got what you mean. Of course, you are right!

I have fixed and pushed that and other issues.

ProLoser:
Does this still support simple keypresses without meta-keys (such as simply hitting the return key?).

In addition, it just occurred to me that you are not providing access to the event object, which although is common, doesn't allow me to preventDefault(). This is something the original implementation was designed to do.

borodiychuk:
Yes, sure. Like `enter`, `esc` or `13`

PR #36, opened 2012-07-08
ui-modal directive

ajoslin:
Wrap of bootstrap modal


ProLoser:
Looks good to me, and since it doesn't inject any new DOM there's no need to do it in `link:{post:fn}`

Now all you have to do is add a demo :)

PR #38, opened 2012-07-09
Jakefile

petebacondarwin:
I have created an initial Jakefile that does everything that the Makefile does.
This requires users to install nodejs, jake and a bunch of node packages.


ProLoser:
Looks like you may have (accidentally) commited all of the node modules.

Think you can add node_modules to the git-ignore and rewrite the commit?

It's fairly easy using git-gui using --append

petebacondarwin:
Hi Dean
I did that on purpose, since they are needed for the build.  From the node
js work that I have done it is quite common to do this since it ensures
that the dependencies that are needed are there:
http://www.mikealrogers.com/posts/nodemodules-in-git.html.  Users of the
project don't deploy the node_modules folder, they are only interested in
the build folder, right?
This is similar to including a Makefile in the project.  It is not needed
by the users but is needed for building the project.
Pete

On 9 July 2012 23:12, Dean Sofer <
reply@reply.github.com

ProLoser:
Why not simply use `npm install`? I mean you've already created (and
assumed) a dependency on Node.js by using jake, I don't see a reason to
include all those packages and clutter the repo (especially for just a
build script) when you could just add 1 line to the instructions that tell
people to install the dependent packages.
I've never seen a workflow where people explicitly add the dependencies,
it's either submodules or a package.json file.

On Mon, Jul 9, 2012 at 10:48 PM, Pete Bacon Darwin <
reply@reply.github.com

petebacondarwin:
Did you read this?
http://www.mikealrogers.com/posts/nodemodules-in-git.html.

On 10 July 2012 07:21, Dean Sofer <
reply@reply.github.com

ProLoser:
Just finished. Was about to add that I plainly disagree with that article<del>. I may go so far as to call him a dumbass</del> for a few reasons:
1. Can't you specify the version of the module to checkout in the package.json? (I am not that familiar with node so I may be wrong)
2. What is the point of a package.json and NPM in general if you simply include everything? What happens if you WANT to keep on track with updates to those repos?

but most importantly
1. If NPM fails to do any of the above, why not simply use git submodules? I know for a FACT that you can not only lock your dependencies into a version, but even a specific commit if you want. No clutter to your extra repo, and if you choose to update those dependencies then it's as simple as changing into that directory and pulling the latest.

ProLoser:
PS: the offense is not intended to you, but towards the author of that article.

ProLoser:
I poked around irc.freenode.net/node.js and asked about the article and community consensus. Apparently that article was written before `package.json` came about and may no longer be relevant.

petebacondarwin:
Not sure if I disagree with him.  I have done a bit of work using NodeJS on
Kanso, and it has definitely been easier to have the node_modules checked
in - it is good to see what you code is being called inside github apart
from anything else and if I wanted to upgrade then I would simply run npm
update and then git commit.
But since this is a community project and not mine I'll add a package.json,
with relevant dependencies and then we can use npm install.
Pete

On 10 July 2012 07:39, Dean Sofer <
reply@reply.github.com

ProLoser:
In github, if you use submodules they will point to other repositories even
on Github, and it will display the commit hash they are referring to so
that you can browse around the precise snapshot your project will be using.
However I believe that the predominant solution nowadays in Node is to rely
(heavily) on package.json. Edward Hotchkiss (who was initially working on
UI with me) complained about my use of tabs because "they didn't look
pretty on Github.com" which... is just sort of a weird sort of logic to me.

On Tue, Jul 10, 2012 at 12:07 AM, Pete Bacon Darwin <
reply@reply.github.com

petebacondarwin:
Hi Dean,

Well, Github is a tool, and a very valuable one at that.  If the use of
tabs makes it difficult to use the tools then I think that is reasonable
logic.  Maybe one could argue that Github should handle tabs better but ...
:-)

Interestingly the npm FAQ has something to say on the topic of including
node_modules in git -  http://npmjs.org/doc/faq.html (look for Should I
check my `node_modules` folder into git?).  It says that if you are
deploying your project (like as a website or an app) then you
_should_commit your node_modules to git.  If you are making a reusable
library then
you should _not_ commit.  In our case, I am not sure which we are doing...
If anything what it might be telling us is that in reality we should
abstract the build process out from the library itself and put it into a
tool of its own.  A bit like bpm (which now seems defunct)??

git submodules are a great tool but probably not the best in this case.  By
putting things in node_modules, nodejs can automatically find them when it
needs them and because they are put there by npm, they have been packaged
up in the correct form for our use, whereas even assuming we could find the
package we need in git, the projects are not always in the form that we
need to consume them.

In the meantime, I am happy to go with package.json and see how thing pan
out.

Pete

On 10 July 2012 08:11, Dean Sofer <
reply@reply.github.com

petebacondarwin:
Ok, so you can now run

```
  npm install
```

to get the node modules that are needed.  And I have tidied it up so that the Jakefile itself is much more readable.

petebacondarwin:
Oh, yes.  And another thing.  I would like to wrap up some of this functionality in real Jake tasks to make it even cleaner but for now this works.

ProLoser:
Is it not in jake tasks? I'll have to poke around.

PR #40, opened 2012-07-09
Prettify Filter

krohrsb:
I have added a filter that will turn strings in the format of common variable naming conventions (underscore, camel, pascal) into something more presentable (space separated first letter capitalized).

home_phoneNumber => Home Phone Number
HomePhoneNumber => Home Phone Number
home_phone_number => Home Phone Number


petebacondarwin:
Thanks BlackBarn.  This seems reasonable.  It would be even better with a few tests!

PR #41, opened 2012-07-10
Ui currency with ui:config options

dandoyon:
Fixed to match current standard of using ui:config options. 
Fixed angular-ui.gh pages, however, this should be pulled before that request.


ProLoser:
We actually shouldn't need the gh-pages branch anymore, that was due to my believing that the assets couldn't be served from the raw url on the demo page. However Andy seems to think otherwise and it appears to work (although chrome throws a warning) and I finally realized how Select2 does it.

Bottom line, we're going to drop the gh-pages branch of angular-ui.

PR #42, opened 2012-07-10
Rewrite uiDate to leverage uiConfig instead of depending on $scope.uiDate

ProLoser:
Pinging @petebacondarwin and/or @dandoyon to review and merge this pull request.

I needed to pass some app-wide defaults to uiDate and noticed it wasn't leveraging our ui.config so I added support for it and removed the dependency on `$scope.uiDate`.

I think we need to completely avoid any assumptions about variables being on the $scope as it clutters people's apps and can cause conflicts. Instead everything should be passed as an expression, and the only thing that you can assume is the ui.config value since we namespace it under our own module and declare it in the module declarations.

I did not see any tests that specifically pertain to $scope.uiDate so I did not make any changes there.


petebacondarwin:
Actually I never intended to rely on the parent scope. It was that naughty Dan Doyon who deleted the line that created an isolated scope and added the uiDate to that scope from the attribute.
Thanks looking at this. I think though that we actually need to observe the attribute rather than a one time val in case the user wants to update it on the fly.

dandoyon:
Oops, hmm, ill need to retrace my steps on that one.

Sorry

/;(

ProLoser:
Hmm good point perhaps I should start switching more directives over to $observe rather than the 1-time $eval.

I ended up merging in this commit because I had a few other people asking about this specific functionality.

PR #45, opened 2012-07-11
CodeMirror directive

boneskull:
Was talking with ProLoser today in #angularjs and he asked if I could add a directive for CodeMirror.  Here it is... seems to work, but I'm still feeling out how to write directives.


ProLoser:
Closing this since I merged in Pete's version. However if you want, it still needs unit tests and there is a bug if the ngModel is not initialized inside the controller as an empty string.

If you check for truthy values before passing it to codeMirror, that is one solution, however falsey values should empty out codemirror. As such I had an if-statement which would pass an empty string, the PROBLEM with this is that it would cause angular to trigger $apply again (since I may be changing the value from 'undefined' to '') and then you get the recursive error.

PR #46, opened 2012-07-11
Codemirror

petebacondarwin:


modules/directives/codemirror/src/codemirror.js
```diff
@@ -0,0 +1,63 @@
...
+                }
+                // Create the new codemirror widget
+                codemirror = CodeMirror.fromTextArea(elm[0], options);
+            };
+
+            // Initialize the code mirror widget
+            updateCodeMirror(uiCodemirrorGet());
+
+            // Now watch to see if the codemirror attribute gets updated
+            scope.$watch(uiCodemirrorGet, updateCodeMirror, true);
```
ProLoser, line 54: Is this better than `scope.$watch(attrs.uiCodemirror, ...` ? Just wondering if it's more efficient to use the $parsed value or the original expression.

ProLoser:
I'm getting this when I tried putting a demo together. See my fork of angular-ui.github.com

TypeError: Cannot call method 'split' of undefined
    at http://localhost/angular-ui.github.com/lib/CodeMirror2/lib/codemirror.js:3188:38
    at Object.setValue (http://localhost/angular-ui.github.com/lib/CodeMirror2/lib/codemirror.js:371:19)
    at Object.setValue (http://localhost/angular-ui.github.com/lib/CodeMirror2/lib/codemirror.js:2033:29)
    at [object Object].$render (http://localhost/angular-ui.github.com/angular-ui/build/angular-ui.js:71:28)
    at Object.fn (http://code.angularjs.org/1.0.1/angular-1.0.1.min.js:134:297)
    at Object.$digest (http://code.angularjs.org/1.0.1/angular-1.0.1.min.js:84:102)
    at Object.$apply (http://code.angularjs.org/1.0.1/angular-1.0.1.min.js:86:58)
    at http://code.angularjs.org/1.0.1/angular-1.0.1.min.js:15:396
    at Object.d [as invoke](http://code.angularjs.org/1.0.1/angular-1.0.1.min.js:26:477)
    at ob (http://code.angularjs.org/1.0.1/angular-1.0.1.min.js:15:317)

petebacondarwin:
I think the problem is that ngModelController initializes the model to NaN, which the CodeMirror widget doesn't like.  A quick fix is to make sure that some controller initializes it better before the directive is rendered.  A proper fix would be to update the directive to handle NaN better.

ProLoser:
On the demo site I initialized the model inside the controller. I tried applying a fix to the directive:

```
ngModel.$render = function() {
if (ngModel.$viewValue) {
  codemirror.setValue(ngModel.$viewValue);
} else {
  codemirror.setValue('');
}
};
```

however this causes a recursive loop of `$apply`

I had too many other things that I wanted to finish so I ended up stabilizing what we had so far, getting a decent demo up and pushing.

I'll open a new bug ticket for this issue.

PR #52, opened 2012-07-13
Refactor to use $timeout instead of $$phase

dandoyon:
This uses angular external api over internal scope.$$phase. It was a bit painful to find secret
sauce for doing unit tests but it works.

### DIFF
--- a/modules/directives/modal/src/modal.js
+++ b/modules/directives/modal/src/modal.js
@@ -1,5 +1,5 @@
 angular.module('ui.directives')
-.directive('uiModal', [function() {
+.directive('uiModal', ['$timeout', function() {
   return {
     require: 'ngModel',
     link: function(scope, elm, attrs, model) {
@@ -7,12 +7,14 @@ angular.module('ui.directives')
           elm.modal(value && 'show' || 'hide');
       });
       elm.on('show.ui', function() {
-        model.$setViewValue(true);
-        if (!scope.$$phase) scope.$apply();
+		$timeout(function() {
+    		 model.$setViewValue(true);
+    	});
       });
       elm.on('hide.ui', function() {
-        model.$setViewValue(false);
-        if (!scope.$$phase) scope.$apply();
+        $timeout(function() {
+        	model.$setViewValue(false);
+        });
       });
     }
   };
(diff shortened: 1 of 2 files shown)


modules/directives/modal/test/modalSpec.js
```diff
@@ -2,18 +2,21 @@ describe('modal', function() {
   var elm, scope;
 
   beforeEach(module('ui.directives'));
-  beforeEach(inject(function($rootScope, $compile) {
+  
+  beforeEach(inject(function($injector, $rootScope, $compile) {
     scope = $rootScope;
+    $timeout = $injector.get('$timeout');
```
ajoslin, line 8: Loks like $timeout gets declared as a global variable here?

ajoslin:
Looks good, except for the global variable thing in the above comment.  This is definitely the much better approach.

dandoyon:
Yeah I know, I thought about that after posting. I tried putting a var $timeout at top of test suite and that does not work. I think it's a bug. So the alternative was to put an inject on each method. Then I thought, well, its not a browser application so should i care?? I'll take another look. I try to follow best practices, sometimes its just painful.

dandoyon:
Am closing this pull request, I didn't include build files and am going to look into making tests not use global $timeout

modules/directives/modal/src/modal.js
```diff
@@ -7,12 +7,14 @@ angular.module('ui.directives')
           elm.modal(value && 'show' || 'hide');
       });
       elm.on('show.ui', function() {
```
dandoyon, line 9: @ajoslin Andy, where are show.ui and hide.ui called from? bootstrap? I'm a little confused.

modules/directives/modal/src/modal.js
```diff
@@ -1,5 +1,5 @@
 angular.module('ui.directives')
-.directive('uiModal', [function() {
+.directive('uiModal', ['$timeout', function() {
```
dandoyon, line 2: I made a boo-boo in not putting $timeout into the function's parameters. This is why I couldn't get spec to work the way I wanted to.

PR #54, opened 2012-07-13
Ui modal

dandoyon:
Fixed the function to reference $timeout and unit test.
I am not including build artifacts, once merge is done it should be built against master and those artifacts
should be pushed.


ajoslin:
Looks good Dan. I guess we'll let dean merge it.

ProLoser:
Man I'm so confused looking at the fork network graph. Did Pete revert a bunch of commits from the main fork?

Anyway, I think anyone can merge in pull requests if it's been reviewed by at least one other person.

PR #65, opened 2012-07-16
So this is intended to let someone pass MULTIPLE parameters

ProLoser:
In order to pass them to the jquery plugin, use ui-options.
Pass the options, comma-separated as you would a function.

Example: `<div ui-jq="accordion" ui-options="'option', optionName , [value]">`

This may be completely unnecessary since the function is only called upon init at this time. RFC


dandoyon:
Something to consider is to have something similar to ui-config with a namespace by plugin.

ProLoser:
That's already there actually. This is just extra. If the ui-config value
is an object and the first parameter is an object then they are merged
together (in the normal fashion). However I figured I _could_ add this for
those usual edge cases if people chose to leverage it, however there is
no... _sensible_ way to reflect this in ui.config. I am considering
removing it altogether and only passing one parameter, but I figured it
didn't really hurt to give people the option.

On Mon, Jul 16, 2012 at 10:24 PM, dandoyon <
reply@reply.github.com

PR #68, opened 2012-07-17
Wordwrap filter

ajoslin:
Have been using this internally a bit, decided to commit it: filter for splitting up large amounts of text into an array by number of chars, optionally joining them with a parameter.

Also included a dependencies.json


ajoslin:
Yeah, so it looks like the use-case I was using this for (a jQuery mobile listview not wrapping the text right) could be fixed by changing one line of the css (Dean helped me with that).

This is almost never useful.. can anyone think of a use-case for this where css wouldn't work?

ajoslin:
Closin' it!

Issue #57, opened 2012-07-14
Event not passed to the Keypress directive


pkozlowski-opensource:
hi!

First of all, thnx for all the work on the angular-ui!

It looks like the original event is not passed into the Keypress directive so it is impossible to call `e.preventDefault()`. The problem is visible on the angular-ui [demo page(http://angular-ui.github.com/#directives-keypress) where pressing Enter moves to a new line in a text editor. Can prepare a jsFiddle with an example if needed to reproduce.

From what I can see the pb was introduced in the c15c6fd7 (with a major refactoring). 

Cheers,
Pawel

ProLoser:
I'm completely aware that it occurs and I figured I would update the documentation to not try e.preventdefault. the issue is that before you would pass the name of a function I could execute and therefore pass whatever I want to. The refactor now let's you simply execute angular expressions (similar to how ng-click and change is done) but since I no longer have direct access to a function I cannot pass custom params to it easily. The challenge them becomes how to solve this elegantly. 

You know what? I just thought of how to solve this. I can set an $event object that people can leverage how they wish. 

pkozlowski-opensource reply@reply.github.com wrote:

hi!

First of all, thnx for all the work on the angular-ui!

It looks like the original event is not passed into the Keypress directive so it is impossible to call `e.preventDefault()`. The problem is visible on the angular-ui [demo page(http://angular-ui.github.com/#directives-keypress) where pressing Enter moves to a new line in a text editor. Can prepare a jsFiddle with an example if needed to reproduce.

From what I can see the pb was introduced in the c15c6fd7 (with a major refactoring). 

Cheers,
Pawel

---

Reply to this email directly or view it on GitHub:
https://github.com/angular-ui/angular-ui/issues/57

pkozlowski-opensource:
Good, exposing $event would be consistent with other angular.js directives (for example http://docs.angularjs.org/api/ng.directive:ngMousedown and other directives of this family).

I guess it would be rather simple to code by I was a bit surprised that I couldn't find test for this directive... Anyway, as I have said, I can help with this one or just have your fix :-)

ProLoser:
Holy crap,  Andy already thought of this and fixes it in #55

ProLoser:
Pinging @borodiychuk for this one. I started converting this to pass the $event object just like uiEvent does, however I ran into a few problems when working with the demo:
1. The event is binded to keydown. It is impossible to preventDefault() on this event.
[...]

PR #73, opened 2012-07-18
Fix for the keypress issue #57 plus tests

pkozlowski-opensource:
Hi! I'm trying to scratch my own itch here and sending a PR with the tests and fix for #57 opened by me for the keypress directive. The commit has mostly tests for the existing functionality plus a small fix making #57 solved for real. This commit also aligns $event handling (and inner working) of the keypress directive with the work done by Andy on the event directive.
I reallllly need this keypress thing working and hate to duplicate efforts on my side so would be great if you could merge this one (or at least comment on all the bad/sub-optimal things I did).


ProLoser:
Awesome. Looks like my syntax was wrong. Thanks for taking the time to do this, it has been on my todolist but not enough time in the day.

Issue #62, opened 2012-07-16
CodeMirror throws an error if model isn't initialized


ProLoser:
CodeMirror apparently does not take kindly to falsey non-string values (NaN, undefined, etc).

I tried the following fix:

```
ngModel.$render = function() {
  if (ngModel.$viewValue) {
    codemirror.setValue(ngModel.$viewValue);
  } else {
    codemirror.setValue('');
  }
};
```

However this actually causes another `$apply()` to fire since the model value is changed from `undefined` to `''` (empty string).

boneskull:
I have a fix for this that I'll submit later today.

PR #76, opened 2012-07-19
fixes issue #62: CodeMirror expects a string

boneskull:
This was my approach to fixing issue #62.  We want to call setValue() basically but we do not want the onChange function to fire, so this avoids that.  I don't know if this is safe, but the test proves it works.

### DIFF
--- a/modules/directives/codemirror/src/codemirror.js
+++ b/modules/directives/codemirror/src/codemirror.js
@@ -53,20 +53,17 @@ angular.module('ui.directives').directive('uiCodemirror', ['ui.config', '$parse'
             // Now watch to see if the codemirror attribute gets updated
             scope.$watch(uiCodemirrorGet, updateCodeMirror, true);
 
+            ngModel.$formatters.push(function(value) {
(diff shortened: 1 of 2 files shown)


boneskull:
This seems to accomplish what we want in a less-evil-hack type of way.  Thanks for the direction.

boneskull:
I want to make sure the type checking is right, and I'm not sure isString() does the job.  What if your `scope.obj = 4`?  I think what we really may be concerned about are if we have null or undefined as the scope value.  I think true/false would be ok, as they will be implicitly converted into "true" and "false".  Empty strings are fine of course.  

Should we also be throwing some sort of an error if your ngModel points to an array or object?

boneskull:
Would it be appropriate then to just throw an Error if we have an object or array?  I am unclear on how we handle errors, and what things should silently fail (if any).

boneskull:
I'm satisfied with this change now, as we are throwing an error if appropriate.  I'll wait a bit and see if anyone objects before merging.

PR #77, opened 2012-07-20
ui-jq Now throws an error if the plugin isn't found

ProLoser:
Thought this would be helpful during development. Am I throwing the error properly? And do I need to return out of the function?

I also removed it from being `{ post: fn() }` since that's mostly pointless

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 node_modules
+*.coffee.js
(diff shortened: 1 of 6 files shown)


boneskull:
And no you don't need to return after you've thrown the error.

ajoslin:
You can also just say expect(functionThatThrows).toThrow(), you don't have to write out the error.

But looks good otherwise, methinks merge.

PR #81, opened 2012-07-21
Moar tests

boneskull:
More/better tests for CodeMirror directive.

### DIFF
--- a/modules/directives/codemirror/test/codemirrorSpec.js
+++ b/modules/directives/codemirror/test/codemirrorSpec.js
@@ -2,16 +2,19 @@
 describe('uiCodemirror', function () {
     'use strict';
 
-    // put some crap in the codemirror options to make sure it's coming through
-    // is there an easier way to do this?
-    angular.module('ui.config').value('ui.config', {codemirror: {bar: 'baz'}});
-
     var scope, $compile;
+    beforeEach(function() {
+        // throw some garbage in the codemirror cfg to be sure it's getting thru to the directive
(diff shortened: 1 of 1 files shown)


ProLoser:
You can keep stacking up commits here too as the build files won't be changed until the version is bumped, and I have a few fixes of my own.

modules/directives/codemirror/test/codemirrorSpec.js
```diff
@@ -1,109 +1,118 @@
...
-    var scope;
+    var scope, $compile;
+    beforeEach(function() {
+        // throw some garbage in the codemirror cfg to be sure it's getting thru to the directive
+        angular.module('ui.config').value('ui.config', {codemirror: {bar: 'baz'}});
+    });
     beforeEach(module('ui'));
-    beforeEach(inject(function ($rootScope) {
-        scope = $rootScope.$new();
+    beforeEach(inject(function (_$rootScope_, _$compile_) {
```
petebacondarwin, line 11: Why are you using _$rootScope_ and _$compile_ here?  Have I missed something clever?

boneskull, line 11: I think you have missed something clever.  I have missed something clever too, because I'm not sure why it works.  At any rate I just wanted to avoid variable name collisions.  I saw this used in other tests so I copied it.

pkozlowski-opensource, line 11: Yup, I saw this dependency with underscores in Andy's event directive first and was wondering how it works... This is a total mystery for now: both how and why it works :-)

PR #82, opened 2012-07-21
Updated grunt to use grunt-coffee and now ignoring built files

petebacondarwin:
This now uses pure nodejs to compile the coffee script.
Coffee script files are built with a .coffee.js extension and these files are ignored by gitignore.
I moved the npm dependencies over to devDependencies in the package.json but on my machine, at least, a straight npm install seems to pull them in anyway.

### DIFF
--- a/grunt.js
+++ b/grunt.js
@@ -92,16 +92,4 @@ module.exports = function(grunt) {
       }
     });
   });
-
- /* grunt.registerTask('coffee', 'compile coffee files', function() {
-    var done = this.async();
-    grunt.utils.spawn({
-      cmd: 'coffee',
-      args: ['-cb', 'modules']
-    }, function(error, result, code) {
-      if (error) grunt.warn(error.stderr);
-      else grunt.log.write(result.stdout);
-      done();
-    });
-  });*/
 };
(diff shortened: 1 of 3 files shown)


ProLoser:
So if we're ignoring the build files we should probably put an extra effort on the build server. My only new concern is how easy would it be to have the build server work with different tags of a repo?

petebacondarwin:
We only need to publish a release when we update the version. When that
happens we manually copy the built files to a github download.
Is this a good interim?

Pete
...from my mobile.
On Jul 21, 2012 8:09 AM, "Dean Sofer" <
reply@reply.github.com>
wrote:

ajoslin:
Looks good.  Although I like having the built files in the repo.  It's worked fine so far to just not commit them.

And also, we don't have the `src` folders anymore on each module, I think you need to update your fork.

petebacondarwin:
Ok. Actually it is kind of nice. I will revert that change.
If we start tagging versions then it doesn't matter if you commit build
files since you can point at a tagged version in things like ghpages.

Pete
...from my mobile.
On Jul 21, 2012 4:11 PM, "Andy Joslin" <
reply@reply.github.com>
wrote:

PR #83, opened 2012-07-21
Adding .idea to .gitignore for people using WebStorm

pkozlowski-opensource:
hi! Just a small PR to add .idea to .gitignore for people using WebStorm

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 node_modules
 *.coffee.js
+.idea


ProLoser:
I kind of suggest adding your local IDE stuff to a [global gitignore](http://jqr.github.com/2009/02/03/global-git-ignore.html) instead.

pkozlowski-opensource:
Oh, I honestly didn't know that global gitignore existed, thnx for the tip! Grr, I guess I just looked at the angular.js .gitignore and saw _plenty_ of ignores in there so I thought it is the only way to go... But it is true that it makes to keep your own IDE setting for yourself. Thnx for the info, did learn something new today!

boneskull:
There may actually be advantages to keeping the .idea folder under version control, such as build configurations and things like that.  (I use PyCharm)

ProLoser:
Ya I figured that might be the case (if you're entire team uses the same
IDE). In that situation a global gitignore may not be an option.

On Sat, Jul 21, 2012 at 12:08 PM, Christopher Hiller <
reply@reply.github.com

PR #84, opened 2012-07-21
Adding tests for the highlight filter

pkozlowski-opensource:
Hi again! Continuing to explore angular-ui codebase and I always find it useful to add tests for the code I'm looking at (of course if a test doesn't exist). So here it is - a PR with the test for the highlight filter.

### DIFF
--- a/modules/filters/highlight/test/highlightSpec.js
+++ b/modules/filters/highlight/test/highlightSpec.js
@@ -1,0 +1,15 @@
+describe('highlight', function() {
+  var highlightFilter, testPhrase = 'prefix highlight suffix';
+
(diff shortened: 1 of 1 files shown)


boneskull:
I'd like to make issues for each of the missing tests and assign them to people. I was going to write a heap of tests too and I don't want to duplicate work.

pkozlowski-opensource:
@boneskull, you are right, we should be opening an issue for missing tests. In fact I was wondering if I should do it before opening a PR but since I had the code ready... Anyway, I agree, let's open an issue for each missing test to avoid duplicating efforts. I'm going to open on for the unique filter.

PR #85, opened 2012-07-21
Renamed Prettify to Inflector

ProLoser:
I also added tests and stubbed out an inflector for underscore and variable.

The tests for those variations currently fail, but I'm leaving them in because I figured this is TDD.


ProLoser:
Also to clarify, I created a 'master' at the top that I figure we will put any changes that haven't been tagged yet. When we tag a new version, we move all the changes inside of 'master' to that version and leave master blank for the next version.

This way we can keep track of changes without forgetting.

ajoslin:
It would be nice to have an option for humanize to capitalize only first letter.  Just a true/false parameter.

ProLoser:
How? 

Andy Joslin reply@reply.github.com wrote:

It would be nice to have an option for humanize to capitalize only first letter. Just a true/false parameter.

---

ajoslin:
Just a guess but should work:
If the option isn't set, do it as is. 

If the option is true, do a non-global replace using the current regex, then do a global replace using one that replaces with lowercase

ProLoser:
I was thinking you meant more along the lines of passing an option to the filter upon execution, but I don't know how to pass more than one parameter. I suppose it might be easier just to make an alternative filter instead.

ajoslin:
Ah, you can do `{{value | filter:param1:param2}}`

PR #88, opened 2012-07-22
NOT TO BE MERGED! Exploratory unit test for the unique filter to be reviewed

pkozlowski-opensource:
hi! Spent some time checking what is wrong with the the unique filter and ended up adding some test. Could you guys have a look at let me know if the assumptions in the tests are correct? As looking at the implementation I don't see how it could possibly work... I mean, given assumptions in my tests the implementation is clearly wrong... I guess I must be missing something obvious here...

Sending this PR to discuss tests / impl, it should NOT TO BE MERGED! as it will break the build (there are failing tests).

### DIFF
--- a/modules/filters/unique/test/uniqueSpec.js
+++ b/modules/filters/unique/test/uniqueSpec.js
@@ -8,26 +8,43 @@ describe('unique', function() {
 
   it('should return unique entries based on object equality', function () {
     var arrayToFilter = [{key: 'value'}, {key: 'value2'}, {key: 'value'}];
-    expect(uniqueFilter(arrayToFilter, '')).toEqual([{key: 'value'}, {key: 'value2'}]);
+    expect(uniqueFilter(arrayToFilter)).toEqual([{key: 'value'}, {key: 'value2'}]);
+  });
+
+  it('should return unique entries based on object equality for complex objects', function () {
+    var arrayToFilter = [{key: 'value', other: 'other1'}, {key: 'value2', other: 'other2'}, {other: 'other1', key: 'value'}];
+    expect(uniqueFilter(arrayToFilter)).toEqual([{key: 'value', other: 'other1'}, {key: 'value2', other: 'other2'}]);
   });
 
   it('should return unique entries based on the key provided', function () {
-    var arrayToFilter = [{key: 'value', other: 'other1'}, {key: 'value2', other: 'other2'}, {key: 'value', other: 'other3'}];
+    var arrayToFilter = [{key: 'value'}, {key: 'value2'}, {key: 'value'}];
     expect(uniqueFilter(arrayToFilter, 'key')).toEqual([{key: 'value'}, {key: 'value2'}]);
   });
 
   it('should return unique entries based on the key provided for complex objects', function () {
-    var arrayToFilter = [{key: 'value'}, {key: 'value2'}, {key: 'value'}];
-    expect(uniqueFilter(arrayToFilter, 'key')).toEqual([{key: 'value'}, {key: 'value2'}]);
+    var arrayToFilter = [{key: 'value', other: 'other1'}, {key: 'value2', other: 'other2'}, {key: 'value', other: 'other3'}];
+    expect(uniqueFilter(arrayToFilter, 'key')).toEqual([{ key : 'value', other : 'other1' }, { key : 'value2', other : 'other2' }]);
+  });
+
+  it('should return unique primitives in arrays', function() {
+    expect(uniqueFilter([1,2,1,3])).toEqual([1,2,3]);
+  });
+
+  it('should work correctly for arrays of mixed elements and object equality', function () {
+    expect(uniqueFilter([1,{key:'value'},1,{key:'value'},2,"string",3])).toEqual([1,{key:'value'},2,"string",3]);
+  });
+
+  it('should work correctly for arrays of mixed elements and a key specified', function () {
+    expect(uniqueFilter([1,{key:'value'},1,{key:'value'},2,"string",3],'key')).toEqual([1,{key:'value'},2,"string",3]);
   });
 
   it('should return unmodified object if not array', function() {
     expect(uniqueFilter('string','someKey')).toEqual('string');
   });
 
-  it('should return unmodified array if key not provided', function() {
+  it('should return unmodified array if provided key === false', function() {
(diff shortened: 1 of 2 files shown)


ajoslin:
Looks good, Pawel.  You should also add a test for arrays of primitives:

``` javascript
it('should return unique primitives in arrays', function() {
  expect(uniqueFilter([1,2,1,3])).toEqual([1,2,3])
});
```

pkozlowski-opensource:
@ajoslin, Andy, thnx for having a look. I agree that it should work for arrays of primitives, adding tests right now. Actually I'm a bit puzzled about the way the filtering key is passed to the filter - from what I could understand from the doc we can have 3 cases:
- [key] {string} the name of the attribute of each object to compare for uniqueness - this makes sense for arrays of objects only
- if the key is empty, the entire object will be compared - will work for both objects and primitives
- if the key === false then no filtering will be performed

Actually the option with the 'empty' key is a bit puzzling - what should be considered as 'empty'? An empty string? or undefined as in your example?

ajoslin:
I guess empty means .length==0 or undefined, yeah.

ajoslin:
And also, it should probably treat a string with no key parameter like an array:

``` javascript
it('should return unique characters in a string, if empty key parameter', function() {
  expect(uniqueFilter('abac')).toEqual('abc')
});
```

ProLoser:
Key was the parameter passed to the filter. If it is undefined, meaning no parameter is passed, it compares entire objects. I realize the undefined vs false case might be weird.

ProLoser:
The string as an array seems like a strange edge case to me.

modules/filters/unique/test/uniqueSpec.js
```diff
@@ -0,0 +1,33 @@
+describe('unique', function() {
+  var uniqueFilter;
+
+  beforeEach(module('ui.filters'));
+  beforeEach(inject(function($filter) {
+    uniqueFilter = $filter('unique');
+  }));
+
+  it('should return unique entries based on object equality', function () {
+    var arrayToFilter = [{key: 'value'}, {key: 'value2'}, {key: 'value'}];
```
petebacondarwin, line 10: What exactly do you mean by "object equality"?  Obviously, here arrayToFilter[0] !== arrayToFilter[2] but some kind of deep equality check might return true.

modules/filters/unique/test/uniqueSpec.js
```diff
@@ -0,0 +1,33 @@
...
+  }));
+
+  it('should return unique entries based on object equality', function () {
+    var arrayToFilter = [{key: 'value'}, {key: 'value2'}, {key: 'value'}];
+    expect(uniqueFilter(arrayToFilter, '')).toEqual([{key: 'value'}, {key: 'value2'}]);
+  });
+
+  it('should return unique entries based on the key provided', function () {
+    var arrayToFilter = [{key: 'value', other: 'other1'}, {key: 'value2', other: 'other2'}, {key: 'value', other: 'other3'}];
+    expect(uniqueFilter(arrayToFilter, 'key')).toEqual([{key: 'value'}, {key: 'value2'}]);
```
petebacondarwin, line 21: I would expect this case to return the whole unique objects from the array not just the matching key and value.

```
expect(uniqueFilter(arrayToFilter, 'key')).toEqual([{key: 'value', other: 'other1'}, {key: 'value2', other: 'other2'}]);
```

modules/filters/unique/test/uniqueSpec.js
```diff
@@ -0,0 +1,33 @@
...
+  });
+
+  it('should return unique entries based on the key provided for complex objects', function () {
+    var arrayToFilter = [{key: 'value'}, {key: 'value2'}, {key: 'value'}];
+    expect(uniqueFilter(arrayToFilter, 'key')).toEqual([{key: 'value'}, {key: 'value2'}]);
```
petebacondarwin: This is not as complex as the previous expectation...

modules/filters/unique/test/uniqueSpec.js
pkozlowski-opensource, line 10: Yes, this is _the_ tricky part.  Depending on the behavior we want to have we can consider 2 objects o1 and o2 being equal if:
- `o1===o2`  - this is reference equality
- `angular.equals(o1, o2)` - this is "deep" equality
- other methods (`angular.toJson(o1)===angular.toJson(o2)` for example - but this is not reliable at all)

I've started implementing equality based on `angular.equals` - this will work fine for all object types but want be super-super fast - O(n*n) - unless anyone knows a good hash-generation function for objects.

modules/filters/unique/test/uniqueSpec.js
pkozlowski-opensource, line 21: yes, you are completely right, since I had many failing tests I didn't catch this obvious thing... Of course the test is wrong here.

modules/filters/unique/test/uniqueSpec.js
pkozlowski-opensource: yeh, good catch once again, I've changed description of the wrong test.... Corrected on my side already but didn't re-submit since started to look into this equality thing (see my previous comments).

modules/filters/unique/test/uniqueSpec.js
ProLoser, line 10: I had no idea about angular.equals()

ajoslin:
(string as array case): Yes, but what else would you as a user expect it to do for `{{'ababc' | unique}}`?  Tests need to cover edge cases too

ProLoser:
I would expect it to do nothing (or fail).

Under that same logic, what do you do if someone does `{{'abadc' | unique:'id'}}`

ajoslin:
There are no repeated instances of 'id', so nothing would be taken out.

pkozlowski-opensource:
OK, pushed a new commit that has both working unit tests and fixes for the filter itself. It will work correctly for arrays but might be not super-efficient for arrays without any duplication.

pkozlowski-opensource:
OK, guys, thank you for all the comments, I've integrated most of them in the b406c1a. The thing I've left for later is filtering on strings - it is not super-difficult to do but since this logic might be confusing let's add support for strings in another commit if needed.

Since I've opened this PR mostly to get you comments I will close this one and will open a fresh one with one commit only so it can be easily merged.

PR #95, opened 2012-07-22
Updated testing to include test-watch-config.js (temporary fix)

dandoyon:
Moved template.js up to match modules structure
Added templateSpec.js and template.js to test configs, they are test ready


dandoyon:
To simplify things I've removed test-watch-config.

PR #96, opened 2012-07-22
Simplified and restructured template, added a filter template using uiConfig. 

dandoyon:
Updated README.md

Before merging, I would like feedback on how this is structured. 
- I introduced a filter with uiConfig but diverged from directives by adding a suffix so filter directive 
  would have ui-config { tmplFilter: { foo : 'foobar' } as opposed to ui-config: { tmpl: { 'foo' : 'bar'}
- I have two examples one for directives one for filters, alongside I provided sample implementations
- I used to have a concept of directive with both two-way and one-way binding, but pulled this out, 
  sometimes too many options can confuse. Also, I wasn't able to convince myself of need for one-way binding.


ProLoser:
Ya, I think it was good to pull out the 1-way vs 2-way cuz I prefer not to use this approach in general and for the reasons you mentioned.

I don't really agree with using a 'Filter' suffix. My thought is not to even bother namespacing directives and filters, but if you guys bite my head off we can use nested config objects, `directives: { select2}, filters: {}` etc. Honestly I'd much rather deal with collisions when we encounter them and not prematurely as it feels like unnecessary clutter ATM.

ProLoser:
I think if these changes are made we can go ahead and merge this in (need to start closing PRs)

dandoyon:
I may pull out the impls entirely and just not confuse. I waffled on adding 'Filter' thought about it
after that it should come out. I'll pull the one-way stuff out of currency filter in a diff branch. 
Thanks for reviewing.  I may get to it tonight or tomorrow.

dandoyon:
I pulled out sample impls and refactored based on feedback, see commits. I think its good to go. 

--dan

Issue #89, opened 2012-07-22
create test for remove directive


boneskull:
Remove needs to be refactored to leverage a comment placeholder. Instead of tracking the index and using elm.data().

boneskull:
this was fixed by PR #98.

PR #98, opened 2012-07-22
#89: added tests for remove directive.  fixed a bug in the remove direct...

boneskull:
...ive where it would not find the sibling nodes because it was detached from the tree.  also fixed a bug where the child variable had no index() function.


ajoslin:
fetched, passed :-) http://d.pr/i/oLQt

PR #99, opened 2012-07-22
ui-if directive with tests

ajoslin:
Taken from https://github.com/tigbro/jquery-mobile-angular-adapter/blob/master/src/main/webapp/utils/if.js


modules/directives/if/if.js
```diff
@@ -0,0 +1,38 @@
...
+ */
+angular.module('ui.directives').directive('uiIf', [function() {
+  return {
+    transclude: 'element',
+    priority: 1000,
+    terminal: true,
+    restrict: 'A',
+    compile: function(element, attr, linker) {
+      return function(scope, iterStartElement, attr) {
+        iterStartElement[0].doNotMove = true;
```
boneskull, line 14: What is this?

ajoslin, line 14: I have no idea.... 
I can't seem to find any docs on it.

ajoslin:
Committed with another repeater, it still works.  I think it _does_ add a comment.

ajoslin:
Is this good to merge?  Should I make a page on angular-ui.github.com for it?

Issue #87, opened 2012-07-22
Fix the unique filter


pkozlowski-opensource:
The unique filter is currently broken and needs to be fixed. On top of this it doesn't have unit tests so those should be added as well.

petebacondarwin:
Oops I was thinking of the unique directive

Pete
...from my mobile.
On Jul 22, 2012 3:28 PM, "Pawel Kozlowski" <
reply@reply.github.com>
wrote:

pkozlowski-opensource:
Fixed in #111.

PR #111, opened 2012-07-24
Fix for the issue #87 plus unit tests

pkozlowski-opensource:
Re-submitting fix for the #87 (after discussion in the initial PR #88). As noted in #88 it is not the most optimized impl on the earth (to do it faster we would need a good hashCode function for js objects) but at least this one works. Please review & merge (or reject) so I can send another PR for the angular-ui web site update.


petebacondarwin:
Looks good to me.

PR #113, opened 2012-07-25
Added ui-required directive with (broken) tests

ProLoser:
I am having trouble trying to figure out how to make the tests properly.

If anyone could lend some assistance with them that would be much appreciated.

The directive is simple and (imho) pretty nifty. Here is a demo implementation: http://plunker.no.de/edit/bHrcd0?live=preview&stream=NgNcNBCHsDeVuOW9

### DIFF
--- a/modules/directives/required/required.js
+++ b/modules/directives/required/required.js
@@ -1,0 +1,22 @@
+/**
+ * Makes an input required based on a condition
+ *
+ * @example <input ui-required="isRequired" ng-model="value">
+ *  		<label><input type="checkbox" ng-model="isRequired"> Is this required?</label> 
+ */
+angular.module('ui.directives').directive('uiRequired', function(){
+	return {
+		require: 'ngModel',
+		link: function(scope, elm, attrs, ngModel) {
+			var required = false;
+			ngModel.$formatters.push(function(value){
+				ngModel.$setValidity('required', !!(required && value));
+				return value;
+			});
+			attrs.$observe('ngRequired', function(value){
+				required = !!value;
+				ngModel.$setValidity('required', !!(required && ngModel.$viewValue));
+			});
+		}
+	};
+});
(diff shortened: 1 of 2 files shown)


ProLoser:
Just realized I should have probably used `toEqual()` instead, however even with this change tests still fail so I will let someone better than I to take a crack at it.

pkozlowski-opensource:
Just out of curiosity, what is wrong with the plain old ngRequire? http://jsfiddle.net/pkozlowski_opensource/t6PGj/8/
I guess I'm missing sth here but it looking at https://github.com/angular/angular.js/issues/861 ng-required should work just fine.... Sorry if couldn't spot a pb with the ngRequired...

petebacondarwin:
I think you are right Pawel.
That being said, I have been playing around with the spec in this commit all evening and I can't get it to work.

ProLoser:
I couldn't find an my required in the docs or API. Mother fucking API. If I seriously wasted all my time I'm going to be annoyed

Pawel Kozlowski reply@reply.github.com wrote:

Just out of curiosity, what is wrong with the plain old ngRequire? http://jsfiddle.net/pkozlowski_opensource/t6PGj/8/;
I guess I'm missing sth here but it looking at https://github.com/angular/angular.js/issues/861; ng-required should work just fine.... Sorry if couldn't spot a pb with the ngRequired...

---

pkozlowski-opensource:
@ProLoser, yeh, it is not documented, the code is here: https://github.com/angular/angular.js/blob/master/src/ng/directive/input.js#L1129; I've bumped into this one in one of the google group threads....

ProLoser:
If there is this built in I can delete this . I would like to know how to get the tests to work though.

pkozlowski-opensource:
@ProLoser, I just had a look at the test and those are fine: I've just changed the `ui-required` to `ng-required` and tests are passing. So in fact it was the directive itself that was problematic, not tests (!).

ProLoser:
Lies and slander!

Well whatever, I deleted the branch anyway.

On Fri, Jul 27, 2012 at 1:03 PM, Pawel Kozlowski <
reply@reply.github.com

PR #117, opened 2012-07-25
Fix for the issue #114. 

pkozlowski-opensource:
The problem was that an empty string passed as a highlight filter was adding a span over each and every character.

### DIFF
--- a/modules/filters/highlight/highlight.js
+++ b/modules/filters/highlight/highlight.js
@@ -5,7 +5,7 @@
  */
 angular.module('ui.filters').filter('highlight', function() {
 	return function(text, filter) {
-    if (!filter) {
+    if (filter === undefined || filter === '') {
 			return text;
 		} else {
 			return text.replace(new RegExp(filter, 'gi'), '<span class="ui-match">$&</span>');
--- a/modules/filters/highlight/highlight.js
+++ b/modules/filters/highlight/highlight.js
@@ -5,7 +5,7 @@
  */
 angular.module('ui.filters').filter('highlight', function() {
 	return function(text, filter) {
-		if (filter === undefined) {
+    if (!filter) {
 			return text;
 		} else {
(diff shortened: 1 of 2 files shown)


cwhistler:
So I updated to the 0.2.0 version and tested again.  I still see issues with bound values.  This code:

```
                   <tr ng-repeat="rec in portfolio.records | filter:query | orderBy:orderProp">
                     <td class="text-left">{{rec.name | highlight:query }}<br/>{{rec.accountTypeDesc}}</td>
                     <td><a href ng-click="goToAcct(rec.id)">Click here to view</a></td>
                     <td class="text-right">{{rec.value | currency}}</td>
                   </tr>

```

results in the following being displayed on the browser.  NOTE this is not the HTML source view, this is actually what is displayed to the user when typing SMITH into the query field:

```
ELZY <span class="ui-match">SMITH</span>
```

The user should not see the span class tags.

ProLoser:
Use the `ng-bind-html-unsafe` directive. I'm not sure if this is something we should be bypassing though as it forces you to allow content you DONT want to be unsafe.

cwhistler:
ok that helped and now this appears to work.  There is an issue if you combine the currency filter with this.  So for example my data is numeric (1000.00) but is displayed with the currency filter ($1,000.00).  In my filter I can type 1000 and the filter is applied correctly, but the highlighting is not.  It stops highlighting at the 1.  I don't know how to resolve this though since the filter is placed at the ng-repeat element, while the highlight is placed on each display element.

ProLoser:
Do you think you could make a fiddle or plunker to help me reproduce the
problem?

pkozlowski-opensource:
@ProLoser, @cwhistler: if I'm not mistaken we can reproduce the issue with this jsFiddle: http://jsfiddle.net/pkozlowski_opensource/8htkH/2/

If this is the correct scenario it is the tricky one since what we get into the highlight filter is already formatted by the currency filter (so we can't use the original filter match). I must say that I don't see any good solution for such use case at the moment...

PR #119, opened 2012-07-27
Introduced ui-sortable directive

teamon:
jQuery UI Sortable plugin wrapper

Example: http://jsfiddle.net/7eTMV/20/ or http://plnkr.co/edit/IcYTPU (jsfiddle is hackish)

I need some help with tests - does anyone know how to start testing drag & drop sorting?

### DIFF
--- a/modules/directives/sortable/REDME.md
+++ b/modules/directives/sortable/REDME.md
@@ -1,0 +1,43 @@
+# ui-sortable directive
+
+This directive allows you to sort array with drag & drop.
+
+## Requirements
+
+- JQuery
+- JQueryUI
+
+## Usage
+
+Load the script file: sortable.js in your application:
+
+    <script type="text/javascript" src="modules/directives/sortable/src/sortable.js"></script>
+
+Add the sortable module as a dependency to your application module:
+
+    var myAppModule = angular.module('MyApp', ['ui.directives.sortable'])
+
+Apply the directive to your form elements:
+
+    <ul ui-sortable ng-model="items">
+      <li ng-repeat="item in items">{{ item }}</li>
+    </ul>
+
+### Options
+
+All the jQueryUI Sortable options can be passed through the directive.
+
+  myAppModule.controller('MyController', function($scope) {
+    $scope.items = ["One", "Two", "Three"];
+
+    $scope.sortableOptions = {
+      update: function(e, ui) { ... },
+      axis: 'x'
+    };
+  });
+
+  <ul ui-sortable="sortableOptions" ng-model="items">
+    <li ng-repeat="item in items">{{ item }}</li>
+  </ul>
+
+
--- a/modules/directives/sortable/sortable.coffee
+++ b/modules/directives/sortable/sortable.coffee
@@ -1,0 +1,46 @@
+###
+ jQuery UI Sortable plugin wrapper
+
+ @param [ui-sortable] {object} Options to pass to $.fn.sortable() merged onto ui.config
+###
+angular.module('ui.directives').directive 'uiSortable', ['ui.config', (uiConfig) ->
+  options = {}
+  if uiConfig.sortable?
+    angular.extend(options, uiConfig.sortable)
+
+  require: '?ngModel'
+  link: (scope, element, attrs, ngModel) ->
+    opts = angular.extend({}, options, scope.$eval(attrs.uiOptions))
+
+    if ngModel?
+      onStart = (e, ui) ->
+        # Save position of dragged item
+        ui.item.data('ui-sortable-start', ui.item.index())
+
+      onUpdate = (e, ui) ->
+        # Fetch saved and current position of dropped element
+        start = ui.item.data('ui-sortable-start')
+        end   = ui.item.index()
+
+        # Reorder array and apply change to scope
+        ngModel.$modelValue.splice(end, 0, ngModel.$modelValue.splice(start, 1)[0]);
+        scope.$apply();
+
+      # If user provided 'start' callback compose it with onStart function
+      _start = opts.start
+      opts.start = (e, ui) ->
+        onStart(e, ui)
+        _start?(e, ui)
+        scope.$apply()
+
+      # If user provided 'update' callback compose it with onUpdate function
+      _update = opts.update
+      opts.update = (e, ui) ->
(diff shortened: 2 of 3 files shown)


ProLoser:
So my coworker was able to create a draggable directive using the
passthrough directive and the event directive has been proving quite
useful. I have been wondering if it's possible to do sortable using these 2
directives together?

I'm a little confused looking at the organization of your code, what is it
doing to enhance integration with angular? Can you comment some of the code
as the logic is a little... opaque to me.

On Fri, Jul 27, 2012 at 11:12 AM, Tymon Tobolski <
reply@reply.github.com

teamon:
I added some comments. Basically this directive hooks at 'update' callback and reorders item in model.

ajoslin:
Couldn't you just do:

``` html
<div ui-jq="sortable" 
  ui-options="mySortableOptions" 
  ui-event="{'update': 'onUpdate($event)', 'receive': 'onReceive($event)'}">
</div>
```

teamon:
Well, you could take the same approach with date picker, yet there is ui-date in angualr-ui.

ProLoser:
@ajoslin I forgot, what was the issue we ran into when we tried the datepicker on the improved passthrough?

ajoslin:
@teamon, well..you have a point :-).  The reason ui-date is a directive really is [this line](https://github.com/angular-ui/angular-ui/blob/master/modules/directives/date/date.coffee#L33).  With the recent changes to ui-jq, the only hard part now is ng-model change -> directive change.

The reasons this could be its own directive are [line 26](https://github.com/angular-ui/angular-ui/pull/119/files#L1R26) and [line 18](https://github.com/angular-ui/angular-ui/pull/119/files#L1R18).

If we could think of a generic way to do this kinda stuff, it would be cool...

ProLoser:
@teamon don't interpret this discussion as negative in any way. Honestly I think we are constantly trying ways to kill off our own directives with more versatile general-purpose directives, otherwise we could easily get caught up developing a new directive for each and every plugin out there.

With that said, I think it would be good to implement this directive.

ajoslin:
I think it would be cool to implement it without the onStart and onUpdate options - people can use ui-event or ui-options for those.  This could be a really minimal directive

ProLoser:
What would it do then if those 2 were left out?

On Fri, Jul 27, 2012 at 2:30 PM, Andy Joslin <
reply@reply.github.com

teamon:
@ProLoser - Don't worry, I don't :) I agree that there should be some general solution. In case of sortable onStart and onUpdate are needed to find out what has changed in the DOM. The other way (notifying DOM about model change) happens ootb in angular.

ajoslin:
Sorry, I mean not having opts.start and opts.update.  People should be pointed to using ui-event or ui-options for these (and the rest of the possible events for ui-sortable), instead of the directive having to handle all of 'em

ProLoser:
If the original plugin takes those as options they should also be
supported, no?

On Fri, Jul 27, 2012 at 2:41 PM, Andy Joslin <
reply@reply.github.com

ajoslin:
http://jqueryui.com/demos/sortable/

Ya, and you can pass them in with ui-options, and in your $scope declare an update function. Eg:

``` javascript
$scope.myOptions = { update: $scope.updateFn }
```

``` html
<div ui-sortable ui-options="myOptions"></div>
```

There are about 15 events that are passed to the element, too, which ui-event can catch.

teamon:
@ajoslin How to "merge" sortable callback (those provided by user and those provided by directive that are required to make the binding work)?

ajoslin:
I guess you would have to compose it.

``` javascript
var options = attrs.uiOptions || {};
function directiveUpdateFn() { ..your current code.. };

if (options.update) {
  var _update = options.update;
  options.update = function(args) {
    _update(args);
    directiveUpdateFn(args)
  }
}
```

But in your directive, I would recommend using $(elm).bind('update') instead of the option, so the users are free to.

teamon:
Ok, I tried with bind("startupdate", function() {.. }) here http://plnkr.co/edit/IcYTPU

There is an issue. Since directive "startupdate" callback and user provided "update" callback are run simultaneously there is indeterministic value of $scope.items (sometimes before drag, sometimes after, sometimes in the middle etc) - not so good.

ajoslin:
I guess you'll have to compose it, then.  It won't end up being far from your original, lol :)

Would you mind adding basic unit tests, too?  There are lots of examples in the other directives.

ProLoser:
I totally need this directive now lol. I have a specific use-case it would help me with.

teamon:
Updated code and example: http://plnkr.co/edit/IcYTPU

I'm working on tests now

teamon:
I've added some specs (really dumb one actually). As stated in comment I can't figure out how to correctly trigger "update" callback so that ng-model could get updated. Any ideas?

ajoslin:
What does the 'ui' object look like?  You could console.log it and create a mock version.

abdulazeezsk:
Hi, change() function is not being fired after the sorting, here is the plunker for the same, http://plnkr.co/edit/PrJNaw?p=preview

Am i missing anything here? Kindly help.

teamon:
`opts.update` is empty, but I'm not sure why

abdulazeezsk:
Looks like scope.$eval is not evaluating in case of isolated scope.

Have posted this on angular group, will update if i get any replies. This would be applicable to all directives(using ui-options) 

Scope.$eval is not evaluating function definition in the case of isolated scope.

http://plnkr.co/edit/PrJNaw?p=preview

Usecase:
Directive(ui-sortable) works fine when it is used directly inside a template. It fires update event once the sorting is completed. http://plnkr.co/edit/xMhQKw?p=preview
Directive(ui-sortable) works fine when it is used directly inside a directive. However, it doesn't fire any event after the sorting. Reason for that is scope.$eval doesn't evaluate the function that is passed as options to ui-sortable directive in case of isolated scope. Here is the plunker for the same, http://plnkr.co/edit/PrJNaw?p=preview

Letsan:
ui-sortable should be added to the docs (no sign of it in http://angular-ui.github.com/).

Most people will indeed use the jQuery passthrough to implement this if they don't know it's there

ghost:
@Letsan :+1:

PR #121, opened 2012-07-28
Removed one-way binding

dandoyon:
No compelling reason to have this.

### DIFF
--- a/modules/directives/currency/README.md
+++ b/modules/directives/currency/README.md
@@ -1,7 +1,7 @@
 # ui-currency directive
 
-In angular, you are able to specify what the currency symbol is (however, you might not want to change it for localization). 
 This directive gives greater control over your currency elements by allowing you to set CSS styles based on the number's sign. 
+In angular, you are only able to specify what the currency symbol is (however, you might not want to change it for localization). 
 						
 ## Usage
 
@@ -11,8 +11,7 @@ Apply the directive to your html elements:
 	    $scope.SomeNumber = 123;
 	});
 
-    <span ui-currency num="SomeNumber"></span> <!-- one way binding -->
-    <span ui-currency ng-model="SomeModel"></span> <!-- two way binding -->
+      <span ui-currency ng-model="SomeNumber"></span>
 
 Default styles are in angular-ui.css and are pretty boring, you could just override these in your
 stylesheet and make things most excellent (e.g. increasing size for negatives, doing a hover sorta thingy )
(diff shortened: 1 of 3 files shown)


petebacondarwin:
One use case I could imagine would look like this:

```
<span ui-currency>{{someNum}}</span>
```

It is nice and clean and seems to me intuitive as to what it does.  Any thoughts?

dandoyon:
I like it! Certainly pretty simple usage. I'll give it a go. Thanks for input.

pkozlowski-opensource:
@petebacondarwin, @dandoyon for me the syntax below:

```
<span ui-currency>{{someNum}}</span>
```

looks like a use-case for a filter, no?

`{{someNum | ui-currency}}`

Maybe we should have both a directive for input fields and a filter for templates? I was using this approach for some custom formatting / parsing functionality (of course reusing code when appropriate).

petebacondarwin:
Sounds good.

Pete
...from my mobile.
On Aug 2, 2012 6:52 AM, "Pawel Kozlowski" <
reply@reply.github.com>
wrote:

dandoyon:
Makes sense, I was thinking that <ui-curr>{{some num}}</ui-curr> would be another

ProLoser:
So what's the conclusion on this? Should I merge it in? As pawel said, read-only seems like a filter to me.

dandoyon:
Not sure about it brings filter because you typically manipulate values whereas this is changing element classes

Go ahead and merge.

Thx 

Dan

ProLoser:
Ya after re-reading the earlier comments I realized you are right.

On Sat, Aug 4, 2012 at 11:19 PM, Dan Doyon <
reply@reply.github.com

> wrote:
> 
> Not sure about it brings filter because you typically manipulate values
> whereas this is changing element classes
> 
> Go ahead and merge.
> 
> Thx
> 
> Dan
> 
> Sent from my iPhone
> 
> On Aug 4, 2012, at 5:39 PM, Dean Sofer reply@reply.github.com wrote:
> 
> > So what's the conclusion on this? Should I merge it in? As pawel said,
> > read-only seems like a filter to me.
> > 
> > ---
> > 
> > Reply to this email directly or view it on GitHub:
> > https://github.com/angular-ui/angular-ui/pull/121#issuecomment-7506694
> 
> ---
> 
> Reply to this email directly or view it on GitHub:
> https://github.com/angular-ui/angular-ui/pull/121#issuecomment-7508020

## 

Dean J Sofer
DeanSofer@gmail.com

Telephone: 714-900-2254
Website: www.DeanSofer.com
Github: github.com/ProLoser

PR #126, opened 2012-07-31
Make grunt.js not have server task, instead use testacular single-run option

ajoslin:
To run tests, you no longer have to start testacular-server.  Testacular uses the single-run option.

The upside of this is no hassle of starting `grunt server` when doing anything.  `grunt` and `grunt watch` handle it all now.

The downside.. it takes about another whole second for the tests to happen, because testacular server has to start itself up each time you're testing now.

What do you guys think?


boneskull:
Is it possible to give us the choice to run the server or not?  I will be running the server because clock time during tests is precious.  The longer it takes to run the tests, the less likely the tests will get run...

pkozlowski-opensource:
@ajoslin, I tend to agree with @boneskull. I don't mind too much starting a server manually, especially that I can easily capture browsers I want to use for testing, while an extra second for each and every test run would be a bit painful. So please, let's make it configurable at least. I guess that if we are going with the automatic-start approach we should also specify browsers to capture in the test config, right?

petebacondarwin:
We should have both options: two test-configs and various grunt tasks.

Pete
...from my mobile.
On Jul 31, 2012 4:22 AM, "Andy Joslin" <
reply@reply.github.com>
wrote:

boneskull:
Andy can you modify the code to allow both methods, or is that not possible?

ajoslin:
Yes, it's possible.  We can have one task do --single-run and the other not.  I don't think we'll need two test-configs.  I'll see if I can get to it tomorrow or this weekend.

ProLoser:
What I was thinking is what if it checked if the server was already running and connected, and if not it started it up?

This way if you want fast tests, just run grunt watch first. If you don't care, just run grunt.

ajoslin:
Nice idea, I'll try it.

ajoslin:
Sorry guys, been super busy lately with moving into my new apartment and work.  I will see if I can do this today...

pkozlowski-opensource:
@ajoslin could you also have a look at my PR #143 as it also changes things around tests running?

ajoslin:
Nice :). Ok I'll put that in too.

I'm not at home now so it'll be later this evening

ajoslin:
After working on this, I can't figure out a way to check if the server is open. Anyone have any ideas?

Maybe we could just make the naked `grunt` command do a single-run test, and maybe even open the browser (Chrome) for the user.  So doing a one-over test is really easy.

Then `grunt watch` will require the server to be open.

PR #132, opened 2012-08-04
Ui validate directive

pkozlowski-opensource:
As discussed on the mailing list this is the first implementation of uiValidate directive (with tests). For now it just supports one validator function and has validation key fixed as 'validator'. There are many extensions possible, notably:
- supporting custom validation keys, 2 syntax proposal below:
  - `ui-validate='myFunction as validationKey'`
  - making it possible for validator functions to return objects like: `{validationKey:true}`
- supporting many validation functions, 2 syntax proposal below:
  - `ui-validate='{validationFunction1 : validationKey1, validationFunction2 : validationKey2}'
  - having one validator function as input but making it possible to return many validation results like `{validationKey1:true, validationKey2:false}`
- supporting any expression as a validator function: `ui-validate='myFunction($value)'`

### DIFF
--- a/modules/directives/validate/test/validateSpec.js
+++ b/modules/directives/validate/test/validateSpec.js
@@ -1,0 +1,99 @@
+describe('uiValidate', function ($compile) {
+  var scope, compileAndDigest;
+
+  beforeEach(module('ui'));
+  beforeEach(inject(function ($rootScope, $compile) {
+    
+    scope = $rootScope.$new();
+    compileAndDigest = function (inputHtml, scope) {
+
+      var inputElm = angular.element(inputHtml);
+      var formElm =  angular.element('<form name="form"></form>');
+      formElm.append(inputElm);
+      $compile(formElm)(scope);
+      scope.$digest();
+
+      return inputElm;
+    };
+  }));
+
+  describe('initial validation', function () {
+
+    it('should mark input as valid if initial model is valid', inject(function () {
+
+      scope.validate = function () {
+        return true
+      };
+      compileAndDigest('<input name="input" ng-model="value" ui-validate="validate">', scope);
+      expect(scope.form.input.$valid).toBeTruthy();
+      expect(scope.form.input.$error).toEqual({validator : false});
+    }));
+
+    it('should mark input as invalid if initial model is invalid', inject(function () {
+
+      scope.validate = function () {
+        return false
+      };
+      compileAndDigest('<input name="input" ng-model="value" ui-validate="validate">', scope);
+      expect(scope.form.input.$valid).toBeFalsy();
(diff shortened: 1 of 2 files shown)


ProLoser:
I am going to be merging everything this weekend so it'd be good if we can get this stuff fixed today or tomorrow.
1. `scope[validateExpr]` is not a good approach because someone may do `$parent.validators.myValidator`. Instead we should probably use `$scope.$eval()` or `$parse`.
2. There is no point to `var validateExpr = attrs.uiValidate || '';` because with uiValidate, the directive wouldn't be executed
3. Support should be added for an object to be passed: `ui-validate="{ size: valSize, shape: valShape }"`. This would allow multiple validation functions to be passed AND each validation rule is namespaced. The keys would then be used as the first argument for `ctrl.$setValidity`

ProLoser:
Please review this commit 79402e63bfd68b40d3a088d7fcdf56b672a77ab6

When I get back from dinner I'll add tests.

pkozlowski-opensource:
@ProLoser, had a look at the commit 79402e63bfd68b40d3a088d7fcdf56b672a77ab6 and it LGTM. I was planning to add support for multiple validators as a next step but since it is nice and easy let's do it now.

@ProLoser, do you want to finish this work yourself? Otherwise I would be happy your changes, add tests etc. and push new commits to the existing PR. I can start working on this in an hour or so. Please let me know so we don't duplicate efforts.

ProLoser:
I haven't added tests yet. Anyone reading this feel free to hop on IRC, as
I will be working on finishing up the next version tonight and tomorrow.
@pkozlowski-opensource I'm cleaning up other directives ATM so any
assistance greatly appreciated.

On Sat, Aug 4, 2012 at 11:51 PM, Pawel Kozlowski <
reply@reply.github.com

> wrote:
> 
> @ProLoser, had a look at the commit
> 79402e63bfd68b40d3a088d7fcdf56b672a77ab6 and it LGTM. I was planning to add
> support for multiple validators as a next step but since it is nice and
> easy let's do it now.
> 
> @ProLoser, do you want to finish this work yourself? Otherwise I would be
> happy your changes, add tests etc. and push new commits to the existing PR.
> I can start working on this in an hour or so. Please let me know so we
> don't duplicate efforts.
> 
> ---
> 
> Reply to this email directly or view it on GitHub:
> https://github.com/angular-ui/angular-ui/pull/132#issuecomment-7508146

## 

Dean J Sofer
DeanSofer@gmail.com

Telephone: 714-900-2254
Website: www.DeanSofer.com
Github: github.com/ProLoser

pkozlowski-opensource:
@ProLoser, took your commit 79402e6 and added tests in the commit 4951b93e11. Both commits are on the `release` branch.

PR #133, opened 2012-08-04
Adding lint task to the grunt build

pkozlowski-opensource:
Just added the lint task so we can start to see what is going on in our code. For now this is just a task def in the build file so people can run it easily at any moment. Later on we can decide how to make it part of the build, decide which rules to enforce etc.

### DIFF
--- a/grunt.js
+++ b/grunt.js
@@ -21,11 +21,11 @@ module.exports = function(grunt) {
     concat: {
       basic: {
         src: ['<banner:meta.banner>', 'common/*.js', 'modules/*/*/*.js'],
-        dest: 'build/<%= pkg.name %>.js',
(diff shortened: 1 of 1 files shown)


ajoslin:
I tried adding it, but so many of our files don't pass.  We need to make it really un-strict, and will have to fix a lot of code to have it work.  But it's probably worth having.

pkozlowski-opensource:
@ajoslin, yep, I saw that many files are not passing and this is why I just added the task, without making it part of the build. This way people can run it, see what is going on and potentially fix some issues. I guess we should start slowly looking at those issues, remove them one by one and when we are in the better shape - make it part of the build. For now I would like to have just a quick way of linting, just to make sure that I'm not introducing new violations.

PR #142, opened 2012-08-10
Fixing failing tinymce tests

pkozlowski-opensource:
Just a small change to the test-config.js to include dependencies on tinymce. Tests were failing due to lack of those dependencies.


boneskull:
My bad, I think I didn't push my changes to test-config.

PR #147, opened 2012-08-13
Inflector: add more (failing) tests and ignore them for now - RFC

pkozlowski-opensource:
Opening this PR mostly to discuss. Today I wanted to quickly fix inflector tests (as I hate where there are failing tests in the build) but realized that there are more problems with this filter. Could you guys have a look and tell me if those tests are OK? If so we could either merge it or add more commits to the issue146 branch.

What is your opinion: should we keep this very simplified version of inflector or replace it with sth more robust? And if so - should we use an external library or write the code ourselfs?


dandoyon:
I agree that failing tests are not good. IMHO, I would take following approach. 
- remove failing tests 
- make sure documentation matches current capability
- put in a feature request for more advanced features for tracking
- add features with more tests

I defer to Dean as to his opinions, since he added the inflector. I did some fixes that didnt get merged, which handled many cases. It was laden with a lot of regex which comes at a cost. Dean suggested we merge two at some point.

PR #153, opened 2012-08-14
Fixes for the reset issue #149

### LINKED ISSUE #149: Broken reset directive
The reset directive doesn't work at http://angular-ui.github.com/

pkozlowski-opensource:
Fixes #149. Updated tests as well to expose the issue (basically input wasn't repaint). Changed the implementation a bit so it works even without jQuery (tests still need jQuery). Fixes #149

### DIFF
--- a/modules/directives/reset/reset.js
+++ b/modules/directives/reset/reset.js
@@ -1,18 +1,21 @@
-
 /**
  * Add a clear button to form inputs to reset their value
  */
-angular.module('ui.directives').directive('uiReset', ['$parse', function($parse) {
+angular.module('ui.directives').directive('uiReset', ['$parse', function ($parse) {
   return {
     require:'ngModel',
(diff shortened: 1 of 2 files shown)


boneskull:
We need to discuss tab length.  I prefer 4, others prefer 2, but I don't want to see a bunch of commits that just clobber other people's lengths.

boneskull:
I see what I missed before in this test; checking the val() of the text field.  Looks good to me.

pkozlowski-opensource:
+1 for discussing formatting. I don't mind either way (spaces, tabs, whatever the length) but it would be good to agree on something. AngularJS uses tabs of 2 AFAIK. Anyway, whatever the decision we should have more or less consistent formatting - otherwise merges get kind of messy (noticed it with this PR).

I will send a quick e-mail to the angular-ui so we can discuss this.

Issue #24, opened 2012-06-20
Changing options in controller doesn't update view


twoolworth:
If you need to change the dateOptions later on based on say a ng-change event from a select it doesn't update the views date picker with the new options. I have a condition that depending on what select option you chose it restricts what days are available to select in the date picker. If I have the restriction in the beginning it works. applying it later within the controllers change handler doesn't work..

My basic controller code:

``` javascript
$scope.startDateOptions = {minDate: 0};

$scope.unavailableDates = [
        "2012/01/29","2012/01/30","2012/01/31",
        "2012/02/29","2012/02/29",
        "2012/03/29","2012/03/30","2012/03/31",
        "2012/04/29","2012/04/30",
        "2012/05/29","2012/05/30","2012/05/31",
        "2012/06/29","2012/06/30",
        "2012/07/29","2012/07/30","2012/07/31",
        "2012/08/29","2012/08/30","2012/08/31",
        "2012/09/29","2012/09/30",
        "2012/10/29","2012/10/30","2012/10/31",
        "2012/11/29","2012/11/30",
        "2012/12/29","2012/12/30","2012/12/31"
    ]; // yyyy/MM/dd

    $scope.unavailableDays = ["Saturday","Sunday"];

    $scope.unavailable = function(date) {
        $scope.ymd = date.getFullYear() + "/" + ("0"+(date.getMonth()+1)).slice(-2) + "/" + ("0"+date.getDate()).slice(-2);
        $scope.day = new Date($scope.ymd).getDay();
        if ($.inArray($scope.ymd, $scope.unavailableDates) < 0 && $.inArray($scope.days[$scope.day], $scope.unavailableDays) < 0) {
            return [true, "enabled"];
        } else {
            return [false,"disabled"];
        }
    }

$scope.selectedFrequencyChanged = function() {
        $scope.showDayOfWeek = angular.equals($scope.selectedFrequency.value, "weekly");

        $scope.startDateOptions = {minDate: 0, beforeShowDay:$scope.unavailable};
    }
```

Here is the select dispatching the change event:

``` javascript
<select id="input_frequency" ng-model="selectedFrequency" ng-options="f.frequency for f in frequencies" ng-change="selectedFrequencyChanged()" class="span2"></select>
```

Here is the input I'm using within the view:

``` javascript
<input ui-date="startDateOptions" id="startdate" type="text" ng-model="service.startdate" value="{{service.startdate}}" class="span3">
```

Am I just missing something?

[...]

PR #155, opened 2012-08-15
Watch date options

petebacondarwin:
This change allows the date directive to $watch the date widget options for changes at run time.
It should fix the problem in issue #24.


ProLoser:
I think you should be using attrs.$observe(), no?

petebacondarwin:
I might be wrong but my understanding is that $observe uses $interpolate
which requires the attribute to be wrapped in {{}}. $watch uses
$eval/$parse which expect a straight expression.

ProLoser:
Crap. Now I have to go double check all my other directives. Have I been
doing it wrong?

On Sat, Aug 18, 2012 at 11:16 PM, Pete Bacon Darwin <
notifications@github.com> wrote:

> I might be wrong but my understanding is that $observe uses $interpolate
> which requires the attribute to be wrapped in {{}}. $watch uses
> $eval/$parse which expect a straight expression.
> On Aug 19, 2012 4:23 AM, "Dean Sofer" notifications@github.com wrote:
> 
> > I think you should be using attrs.$observe(), no?
> > 
> > —
> > Reply to this email directly or view it on GitHub<
> > https://github.com/angular-ui/angular-ui/pull/155#issuecomment-7850267>.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/angular-ui/angular-ui/pull/155#issuecomment-7850864.

## 

Dean J Sofer
DeanSofer@gmail.com

Telephone: 714-900-2254
Website: www.DeanSofer.com
Github: github.com/ProLoser

pkozlowski-opensource:
yep, @petebacondarwin is right, interpolate is for values containing `{{}}`, sth that href is using for example: `src="mypath/{{foo}}/fixedpart/{{bar}}"`.

So in the end it all boils down to supporting multiple variables in an attribute vs. attribute being an expression.

petebacondarwin:
Also $interpolate returns a string not an object

Pete
...from my mobile.

PR #156, opened 2012-08-16
Added a new Extract and Format filter with test

ProLoser:
Extract returns a new array of strings based on the paths. Ideally paths will allow us to traverse more deeply than one layer, but for now it's only 1 level deep. You can pass a separator and an array of paths to merge together multiple properties of the object. Useful for an array of people and you want to create a list of firrstName lastName, etc.

It returns an array which you can wrap and call join() on. Example:

``` html
<pre ng-bind-html-unsafe="(data|extract:['name','age']:': ').join('\n')"></pre>
```

Format is an i18n utility that I use that I actually put on the string prototype. Half of our filters I wonder might perhaps be better suited on the primitive prototypes. Anyway, it lets you swap out tokens from the string. It's very flexible and powerful. Example:

``` html
<pre>{{'Hello :name, how is the :object :day'|format:{name:'Bob',object:'sun',day:'tomorrow'} }}</pre>
```

### DIFF
--- a/modules/filters/extract/extract.js
+++ b/modules/filters/extract/extract.js
@@ -1,0 +1,38 @@
+
+/**
+ * Extracts properties of an array of objects and returns an array of strings
+ *
+ * @TODO Make paths go deeper than 1 level
+ * @TODO Allow an array for separator param?
+ * @param dataset {array[object]} array of objects to extract from
+ * @param paths {mixed} the property to retrieve from the objects
+ *   string: property name
+ *   array[string]: array of property names to be joined by separator in the order passed
+ * @param [separator] {string} Optional the string to join multiple properties
+ */
+angular.module('ui.filters').filter('extract', function(){
+  // paths can optionally be an array
+  // separator is how to merge those different paths
+  return function(value, paths, separator) {
+    var result = [],
+        newItem;
+    if (angular.isArray(value) || angular.isObject(value)) {
+      angular.forEach(value, function(item) {
+        // if paths is an array
+        if (angular.isArray(paths)) {
+          newItem = [];
+          angular.forEach(paths, function(path){
+            newItem.push(item[path]);
+          });
+          result.push(newItem.join(separator));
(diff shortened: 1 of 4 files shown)


ProLoser:
Just realized I forgot to do the extract filter specs.

ajoslin:
All looks good, although your extract is usually called `pluck` in javascript - see `_.pluck`: http://underscorejs.org/#pluck

ajoslin:
Actually, maybe we should just not remake underscore's stuff.. People tend to just use underscore if they want helpers, and they can just assign `$rootScope._` to use it. Thoughts?

pkozlowski-opensource:
+1 for not re-writing underscore if we can avoid it....

ProLoser:
I spose. I don't know if underscore can pull out multiple properties and
merge them together (like firstName and lastName). I just thought I would
share something I use internally. We don't have to merge it in. What about
the format method?

On Fri, Aug 17, 2012 at 12:36 PM, Pawel Kozlowski
notifications@github.comwrote:

> +1 for not re-writing underscore if we can avoid it....
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/angular-ui/angular-ui/pull/156#issuecomment-7832774.

## 

Dean J Sofer
DeanSofer@gmail.com

Telephone: 714-900-2254
Website: www.DeanSofer.com
Github: github.com/ProLoser

pkozlowski-opensource:
Had a look at this PR and I'm not sure we should be merging the extract filter (btw most of tests are failing for it....).

As for the format filter I would be in favor of merging it. Had a quick look at the code and there are 2 tests failing... Quickly fixed them in cff71e091d (pr156 branch) but I guess the best would be to open a separate PR with one filter only and a fixed test. BTW: lint doesn't complain :-)

ProLoser:
Lint doesn't complain about my code :-) lol.

I forgot to do the tests for one of the filters so it's just placeholders.

Issue #129, opened 2012-08-02
bootstrap-transitions seems to break modal directive


rosslavery:
I'm not certain of why it's happening, but it seems as though adding bootstrap-transitions.js (required for fading in/out of modals) is incompatible with the directive written for angular-ui.

If both bootstrap-modal and bootstrap-transition are included, the "in" class is never added to the modal element, and thus never shown. The overlay element is added to the DOM, but upon pressing escape, it is only hidden (via opacity), but not removed from the DOM as it should be. This prevents clicking on any page element after that point.

It seems reproducible to me, I can comment in/out the bootstrap-transition file, and it works and then doesn't.

I've tried to narrow it down and it seems like there's an issue with line 56 of http://twitter.github.com/bootstrap/assets/js/bootstrap-modal.js

The callback for backdrop.call doesn't seem to get executed, that's about as far as I can get with my skillset unfortunately.

ajoslin:
Hi rabhw, thanks. Sorry for the late reply. 

I will look at this tonight.

rosslavery:
Hey Andy,

Sorry to pester, just wondering if you've had a chance to take a look at what's causing this.

ajoslin:
Never did look.. I'll check now.

rosslavery:
Thanks a lot Andy, appreciate it.

ajoslin:
If you copy/paste the pull request into your angular-ui.js for now, it should fix it.  Let me know, I tested it on my own and it seems fine now.

PR #159, opened 2012-08-17
Fix ui-modal when transitions are enabled

ajoslin:
Address issue #129


ProLoser:
What happens if jquery is undefined?

ajoslin:
The whole bootstrap-modal.js depends on jQuery already.

ProLoser:
Touche salesman

PR #160, opened 2012-08-17
Grunt custom-builder

ajoslin:
With this commit, you can build a custom angular-ui.js to build/custom/angular-ui.js by doing this:

``` sh
grunt build:directives/modal:directives/event:filters/inflector
```

The next step is to hook up this task to the rest of the build process (so you just give the current build task a list of modules and it does it as normal, into your custom folder).

### DIFF
--- a/grunt.js
+++ b/grunt.js
@@ -3,6 +3,14 @@ var testacular = require('testacular');
 /*global module:false*/
 module.exports = function(grunt) {
 
+  var modules = {};
+  grunt.file.recurse('modules', function(abspath, rootdir, subdir, filename) {
+    //If numbers / is 1, we know it's first in depth. eg filters/inflector, not filters/inflector/test
+    if (subdir.split('/').length == 2) {
+      modules[subdir] = grunt.file.expand(rootdir+'/'+subdir+'/*.js');
+    }
+  });
+
   grunt.loadNpmTasks('grunt-recess');
   grunt.loadNpmTasks('grunt-coffee');
 
@@ -62,7 +70,7 @@ module.exports = function(grunt) {
     },
     watch: {
       files: ['modules/**/*.coffee', 'modules/**/*.js', 'common/**/*.js', 'templates/**/*.js'],
-      tasks: 'coffee test'
+      tasks: 'coffee concat:basic test'
     }
   });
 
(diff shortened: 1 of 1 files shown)


dandoyon:
cool

grunt.js
```diff
@@ -3,6 +3,14 @@ var testacular = require('testacular');
 /*global module:false*/
 module.exports = function(grunt) {
 
+  var modules = {};
+  grunt.file.recurse('modules', function(abspath, rootdir, subdir, filename) {
+    //If numbers / is 1, we know it's first in depth. eg filters/inflector, not filters/inflector/test
+    if (subdir.split('/').length == 2) {
```
boneskull: If you're concerned about jshint/jslint it is going to complain about the ==

grunt.js
```diff
@@ -95,4 +103,14 @@ module.exports = function(grunt) {
...
   });
-};
\ No newline at end of file
+
+  grunt.registerTask('build', 'Build a custom angular-ui.js', function() {
+    var files = ['common/module.js'];
+
+    this.args.forEach(function(moduleName) {
+      if (modules[moduleName])
+        files = files.concat(modules[moduleName]);
```
boneskull: Likewise it will complain about a lack of curly braces

boneskull:
This looks good.  Are you going to wait to merge the PR until you complete the rest of the task?

ajoslin:
Didn't wanna merge it until it actually builds properly.  Right now it does its own little concat task, but it needs to work like the rest of the build process for the chosen modules.

ajoslin:
I hooked it up to the rest of the build process. It's really cool, there's a "build" task now which, when given 0 args, builds everything like it used to, and when given args builds the given modules.

You can do `grunt build:directives/directive1:directives/directive2:filters/filter1` and it will build all the less/js files just from those modules into build/custom.

ajoslin:
Now a few questions, some dean brought up.
1. How can we cache the main build files so they aren't re-built each time?  I was thinking a redis-store on the main server.
2. Should we cache custom builds for each version?  Is that necessary? Another redis-store would work
3. Should we maybe change the header of a custom build?

ProLoser:
We should go over those issues on the build server.

PR #164, opened 2012-08-18
PR 161: Fixing the datepicker timezone issue

boneskull:
OK, here's another go at it.  I added a unit test.  I thought I had broken other stuff.


petebacondarwin:
Hi Christopher

You are probably right that this is not needed but it was added because of
an issue someone had with the date widget parsing strings.
I can't remember exactly what now.

You shouldn't have that comment about what you changed in the code. That is
what the commit message is for.

Also I am not sure what the new unit test is checking.

Pete
...from my mobile.

PR #172, opened 2012-08-20
Added Calendar Directive

joshkurz:
Added in a directive for fullcalendar.js created by https://github.com/arshaw/fullcalendar.

The first commit was a test run. I would only focus on the second commit. The third commit is just a file deletion. 

Changed the directive to pull in events from the ngModelController.
Got rid of the template.
Made the calendar more generic for the user.

Here is a working version of the directive. http://jsfiddle.net/FYn9z/4/

### DIFF
--- a/modules/directives/calendar/calendar.js
+++ b/modules/directives/calendar/calendar.js
@@ -36,13 +36,10 @@ angular.module('ui.directives').directive('uiCalendar',['ui.config', '$parse', f
             eventMouseover: function(event, jsEvent, view) {
             if (view.name !== 'agendaDay') {
               $(jsEvent.target).attr('title', event.title);
-<<<<<<< HEAD
              }
             },
-=======
             }
           },
->>>>>>> 41b260c1ef4461c32c957fd5e2a60e1a304c69e0
         
             // Calling the events from the scope through the ng-model binding attribute. 
             events: ngModel(scope)
--- a/modules/directives/calendar/calendar.js
+++ b/modules/directives/calendar/calendar.js
@@ -1,56 +1,65 @@
-/*
+    /*
 *  Implementation of JQuery FullCalendar 
 *  inspired by http://arshaw.com/fullcalendar/ 
 *  
-*  Basic Calendar Directive that takes in live events as the ng-model and then calls fullCalendar(options) to render the events correctly. 
+*  Basic Angular Calendar Directive that takes in live events as the ng-model and watches that event array for changes, to update the view accordingly. 
 *  
+* Authors
+*  @andyjoslin
 *  @joshkurz
 */
 
 angular.module('ui.directives').directive('uiCalendar',['ui.config', '$parse', function (uiConfig,$parse) {
 
-    uiConfig.uiCalendar = uiConfig.uiCalendar || {};              
-    //returns the calendar
+    uiConfig.uiCalendar = uiConfig.uiCalendar || {};    
+    //returns the calendar     
     return {
-      require: 'ngModel',
-      restrict : "A",
-      replace : true,
-      transclude : true,
-      scope: {
-        events: "=ngModel"
-      },
-       
-
-    link: function (scope, elm, $attrs, ngModel) {
-      var expression,
-        options = {
-          header: {
-            left: 'prev,next today',
-            center: 'title',
-            right: 'month,agendaWeek,agendaDay'
-          },
-
-          // add event name to title attribute on mouseover. 
-          eventMouseover: function(event, jsEvent, view) {
+        require: 'ngModel',
+        restrict: 'A',
+        scope: {
+          eventChanged: "=changed",
+          events: "=ngModel"
+        },
+        link: function(scope, elm, $attrs) {
+            var ngModel = $parse($attrs.ngModel);
+            //update method that is called on load and whenever the events array is changed. 
+            function update() {
+            //Default View Options
+            var expression,
+              options = {
+                header: {
+                left: 'prev,next today',
+                center: 'title',
+                right: 'month,agendaWeek,agendaDay'
+              },
+            // add event name to title attribute on mouseover. Would be nice if this was an angular popover. 
+            eventMouseover: function(event, jsEvent, view) {
             if (view.name !== 'agendaDay') {
-              (jsEvent.target).attr('title', event.title);
-            }
-          },
+              $(jsEvent.target).attr('title', event.title);
+             }
+            },
         
-          // Calling the events from the scope through the generic ng-model binding attribute. 
-          events: scope.events
-        };
-
-      if ($attrs.uiCalendar) {
-        expression = scope.$eval($attrs.uiCalendar);
-      } else {
-        expression = {};
-      }
-
-      angular.extend(options, uiConfig.uiCalendar, expression);
-      //use the options object to create the personalized calendar
-      elm.fullCalendar(options);
-    
-    }
-  };
+            // Calling the events from the scope through the ng-model binding attribute. 
+            events: ngModel(scope)
+            };          
+            //if attrs have been entered to directive, then create a relative expression. 
+            if ($attrs.uiCalendar)
+              expression = scope.$eval($attrs.uiCalendar);
+            else 
+              expression = {};
+              //extend the options to suite the custom directive.
+              angular.extend(options, uiConfig.uiCalendar, expression);
+              //call fullCalendar from an empty html tag, to keep angular happy.
+              elm.html('').fullCalendar(options);
+            }
+            //on load update call.
+            update();
+            //watch for changes to the eventChanged object passed into the directive, and update if changed. 
+            scope.$watch(function() {
+                return scope.eventChanged;
+            }, function() {
+                update();
+            },true);
+        }
+    };
 }]);
--- a/modules/directives/calendar/calendar.js
+++ b/modules/directives/calendar/calendar.js
@@ -38,8 +38,6 @@ angular.module('ui.directives').directive('uiCalendar',['ui.config', '$parse', f
               $(jsEvent.target).attr('title', event.title);
              }
             },
-            }
-          },
         
             // Calling the events from the scope through the ng-model binding attribute. 
             events: ngModel(scope)
--- a/modules/directives/calendar/test/calendarSpec.js
+++ b/modules/directives/calendar/test/calendarSpec.js
@@ -23,7 +23,31 @@ describe('uiCalendar', function () {
     //Creates a calendar with with two expressions height and weekends
     function createCalendar0(events) {
       scope.events = events || {};
-      $compile("<div ui-calendar='{height: 200, weekends: false}' ng-model='events'></div>")(scope);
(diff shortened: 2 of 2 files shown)


petebacondarwin:
Nice one Josh. I think Dean has picked up on most things. It is great that
you included tests!

I guess you are not that used to git? To remove a folder/file you need to
do git rm then a git commit rather than just deleting the local file.

Why have you committed changes to test-config.js?

By the way I thought we were not including library dependencies directly in
the project?

It would also be great if you could learn about git rebase as this would
have allowed you to squash some of your commits before pushing.

Pete
...from my mobile.

pkozlowski-opensource:
Pulled this PR into my repo and the directive looks cool and tests are passing in chrome. Still needs some tidying up:
- see my in-lined comment in the code
- you need to get rid of the fullcalendar folders (`git rm -r modules\directives\fullcalendar`, `git commit -m "Removing fullcalendar"`)
- lint is complaining both about a directive and tests: run `grunt lint` to see what is going on

Would be great if you could push another commit with the corrections identified so far, so we could do the final review.

joshkurz:
Yes this is my first real experience working with a group on git hub, so I'm learning right now. 
Thanks for the help and the patience. 

I will add these changes to the repo later tonight.

joshkurz:
OK so I just pushed a new version of the directive with the additions. 

grunt-lint is very nice. I have fixed all of the errors that were discovered by this tool except for one. Not sure why the output says angular is undefined on line 48 of calendar.js. I am using angular.extend to fetch the attrs from the linked directive. Other directives inside angular-ui utilize this same approach. I am kind of puzzled about this actually.

petebacondarwin:
I suspect that it is just telling you that you are relying on something
else adding angular to the global namespace.

Pete
...from my mobile.

petebacondarwin:
By the way I have been using a great plugin to sublime text 2 that runs
jslint in the background on the fly and effectively prevents you from
writing bad js in real time. The grunt task is then just a safety net.

Pete
...from my mobile.

pkozlowski-opensource:
@joshkurz Josh, lint reporting angular not being defined is due to `'use strict';`, other directives are not using this.

joshkurz:
Oh ok that makes since. Well then I will push another commit with the most
recent changes in place.

Josh Kurz

joshkurz:
Just pushed a new version of the calendar directive. I Added 3 new tests which all pass. I wanted to make sure the user could add in fields to the uiCalendar attribute to ensure customization. 

Here is a link to the directive with attributes inserted. http://jsfiddle.net/FYn9z/6/

pkozlowski-opensource:
@joshkurz I'm sorry but I've found another problem that is possible to reproduce with your jsFiddle. So basically putting a mouse over an event gives error in the console, for example Chrome says:

Uncaught TypeError: Object #<HTMLDivElement> has no method 'attr' angular-ui.js:50
options.eventMouseover angular-ui.js:50
T fullcalendar.min.js:22
f fullcalendar.min.js:90
(anonymous function) fullcalendar.min.js:91
jQuery.event.special.(anonymous function).handle jquery-1.7.1.js:3535
jQuery.event.dispatch jquery-1.7.1.js:3256
elemData.handle.eventHandle jquery-1.7.1.js:2875
jQuery.event.trigger jquery-1.7.1.js:3144
(anonymous function) jquery-1.7.1.js:3781
jQuery.extend.each jquery-1.7.1.js:658
jQuery.fn.jQuery.each jquery-1.7.1.js:271
jQuery.fn.extend.trigger jquery-1.7.1.js:3780
(anonymous function) fullcalendar.min.js:42
jQuery.event.dispatch jquery-1.7.1.js:3256
elemData.handle.eventHandle

It is a pity since it is a nice directive but it simply must work...

ProLoser:
Ignore Travis, he still isn't working

joshkurz:
Just updated. That was a really tiny problem. I added $(jsEvent.target) to
fix the problem.

But just when I thought everything was gravy, the Travis Bot threw this
error on the thread. Not sure why.

## 

_Josh Kurz_
*
*

pkozlowski-opensource:
@joshkurz just pulled your changes and mouse errors are gone. Lint is not complaining, tests are passing, directive works so it all looks good. Still need to figure out all the details of this directive but for now it LGTM. Thanks for your work on this!

Could you prepare a PR for documentation as well? It is another repository: https://github.com/angular-ui/angular-ui.github.com

Ignore Travis, ProLoser is just setting it up.

joshkurz:
Yeah sure no problem.

## 

_Josh Kurz_
*
*

ephillipe:
Hi all,
I trying to use AngularUI Directive for FullCalendar, but I need help.
The code work only if I set Events on controller load. If I add Events later, the calendar is not updated. 

I create a Fiddle for exemplification.  http://jsfiddle.net/ephillipe/DXDpG/

After click in Add child button, see how itens are added and update UL element, but no fire $watch function.

Maybe I doing some wrong?

joshkurz:
No your doing anything wrong. There is no fire watch method in place. There
should be though.

Josh Kurz

ephillipe:
I change the Fiddle. 
If I point $watch for integer property, the $watch is fired. If I point for Array, do not work.

Any idea?

joshkurz:
We need to add the watch function inside the directive. The directive
should update itself if the scope.events object is changed. We should not
need to watch anything inside of the controller to update the directive, it
should all be done via the calendar directive. So its a good thing you
created this use case.

here is a fiddle where I have extracted the calendar, for easy development. If anyone has any ideas please chime in. http://jsfiddle.net/xqjtw/1/

Thanks
Josh Kurz

joshkurz:
Ok so @AndyJoslin connected the calendar to watch the events from inside the directive depending on the events' array being altered. I hacked a way to do this without looking at the length. Its actually very odd, because if the directive watches the event model object for changes, nothing happens, but if it watches the length or some number that is altered on change, it works. We need to look at something other than the length, because an event could just be altered and then angular would still need to bind the changes to the view. 

I wired an inline editor into the ng-repeat to show that we can also change the title of the events as well.
However this is not firing as soon as the title is changed. The title can be changed if an event is added, removed, or
any navigation button is hit. Something has to happen to the scope to see the change. If someone could help with this error it would be much appreciated. 

Also wired up the jquery-ui editable attribute. so now all events can be dragged and dropped. 

There is some console.log debug info going on. Im trying to edit the eventChanged variable from both directives to update the view as soon as the inline editor is blured. 
http://jsfiddle.net/joshkurz/xqjtw/43/

petebacondarwin:
Can we get better error messages from the Travis bot?

Pete
...from my mobile.

pkozlowski-opensource:
AFAIK Travis is still not well configured so I wouldn't worry about messages atm. We need to get it working first.

joshkurz:
I was able to do some hacking and get the model to update on scope.event changing, but It was not good enough. I was calling apply from inside the inline editor to update the scope, which was having trouble to say the least. Anyways it would not even matter if it wasn't, because scope.$watch should fire from inside the calendar directive on any change to the events array. 

Currently the configuration works with the calendar directive watching event.length, but only if an event is added or removed from the event array. I have installed an inline editor directive to alter the state of any event object title. If the title is updated, then the scope is not updated automatically, because we are only watching the length. We need to watch the events array objects themselves for changes, but when this is tried a digest error "Error: 10 $digest() iterations reached. Aborting!" appears in the console on every change. There should be no reason that we cant watch events for any changes. 

I have read in related threads that watching entire arrays is a bad idea, because the arrays could grow out of proportion and con-volute the digest cycle. So with this in mind I tried to edit the actual fullcalendar.js file and add in the correct angular bindings so the title would be attached to the scope. This failed, because the calendar is loaded after angular is rendered. 

I'm thinking i'm doing something wrong or this is a bug of some kind. I understand that its bad practice to watch an entire array for changes, but in this case I think it might be the only way to get this done properly.

http://jsfiddle.net/joshkurz/xqjtw/52/  -  Here is the version watching events.length
http://jsfiddle.net/joshkurz/xqjtw/51/  - version watching events

Here is where I tried to add the binding from inside fullcalendar.js.  
http://jsfiddle.net/joshkurz/xqjtw/54/ - failed

tkrotoff:
Hi all,

I hope that I'm wrong on this but I don't think it is interesting to have a directive for FullCalendar, here is why:

FullCalendar is smart enough to [only fetch the events that he needs](http://arshaw.com/fullcalendar/docs/event_data/lazyFetching/).
Example: FullCalendar will only fetch from the server (using AJAX) the events from march 2012 if the user asks for march 2012 (by clicking on next/prev for example).

By using a FullCalendar directive, then you need a controller that will fetch all the events from day one and save them inside $scope.events. If you have 100000 events in database, $scope.events will contain 100000 events => a lot of RAM + network time.

The FullCalendar directive has no way to know how to fetch the minimum amount of events FullCalendar needs.

ProLoser:
So you're saying DONT add a fullcalendar directive? What stops you from leveraging the fullcalendar's native ability in this directive? I haven't had an opportunity to dive into the code for this yet (still homeless) so I don't have all the details honestly

tkrotoff:
> So you're saying DONT add a fullcalendar directive?

Well, I was too fast writing my previous post :) To have a directive for FullCalendar is nice for some syntactic sugar:

``` HTML
<div ui-calendar></div>
```

instead of:

``` JavaScript
$('#calendar').fullCalendar({
  // ...
});
```

> What stops you from leveraging the fullcalendar's native ability in this directive?

Nothing, you are right.

Current directive implementation uses [FullCalendar events as an array](http://arshaw.com/fullcalendar/docs/event_data/events_array/).
Instead it should use (or at least give the possibility to choose) [FullCalendar events as a json feed](http://arshaw.com/fullcalendar/docs/event_data/events_json_feed/) and/or [FullCalendar sourceEvents](http://arshaw.com/fullcalendar/docs/event_data/eventSources/) in order to fetch the minimum amount of events FullCalendar needs.

To sum up, the directive implementation should not impose nor suggest a controller that stores all the events ($scope.events).

ProLoser:
I've been too busy to review the code thoroughly but none of our directives impose this sort of behavior and as such neither will FullCalendar when it is merged :)

I could have sworn however that the current version leverages ui-model. However, if you leave out the data and instead pass in the options to populate the json-related dataset, does this function properly?

joshkurz:
Ok so tkrotoff you are right the calendar needs to be able to pull in a live source as an event feed.  ProLoser you are right as well, we are utilizing the ui-model, so we can just place in a url string to the feed and utilize the fullcalendar's native functionality or should be able to. 

here is a fiddle that shows this in theory, but it does not work because jsfiddle does not allow http request to go through. I will create a better example of this. 

http://jsfiddle.net/joshkurz/xqjtw/57/

joshkurz:
Ok nevermind it was not a jsfiddle problem I was not including the correct gcal.js file in the example. Here is a working example with the live event feed from a random google calendar. 

http://jsfiddle.net/joshkurz/xqjtw/59/

PR #176, opened 2012-08-25
bug 161: fixes timezone problems in date picker.  implements changes sub...

boneskull:
...mitted by Pete. (removed Date object conversion from tests and directive as these were causing timezone issues)


boneskull:
couldn't automatically merge... need to rebase or something.

petebacondarwin:
Have a look at the network in github. Your branch is probably behind the
master. Now you need to pull in that and merge then manually push to the
reps.

Pete
...from my mobile.

PR #177, opened 2012-08-25
bug 161: fixes timezone problems in date picker.  implements changes sub...

boneskull:
...mitted by Pete. (removed Date object conversion from tests and directive as these were causing timezone issues)


boneskull:
THERE we go

boneskull:
OK, jslint/jshint should be more happy with this.  Go ahead and merge if it looks OK to whomever.

boneskull:
Pete's changes fixed issue 167, but I just added a test to show that.

pkozlowski-opensource:
Got one test failing as well:

Chrome 23.0 uiDate should parse the date correctly from a string FAILED
        Expected '2012-08-16' to equal '2012-08-17'.
        Error: Expected '2012-08-16' to equal '2012-08-17'.
            at D:/work/gitrepos/angular-ui/angular-ui/modules/directives/date/test/dateSpec.js:48:83

I guess it has sth to do with timezones and string to Date parsing. I'm kind of uncomfortable with the fact that we do support sates as string in the model. I thought that we expect users to provide Data object (and going to bind a Date object). But those are just my 3 cents....

pkozlowski-opensource:
Otherwise LGTM

petebacondarwin:
There are definitely users who want to be able to pass in a string. We
accept that but convert it to a date object that overrides the string in
the scope.

... sent from my tablet

pkozlowski-opensource:
@petebacondarwin, OK, I see. I assume that the value returned from the date picker will be the Data object and it will override the initial string, right?

petebacondarwin:
Yep

... sent from my tablet

Issue #180, opened 2012-08-27
grunt build doesn't work on windows


pkozlowski-opensource:
After changes introduced in the #160 `grunt build` doesn't work on windows anymore (it produces almost empty build files). This is due to hard-coded unix-style paths: https://github.com/angular-ui/angular-ui/blob/master/grunt.js#L76

The remedy is simple, we should be using `path.sep` from node's path. PR will follow shortly.

petebacondarwin:
Actually it would be nicer to replace the grunt.file.recurse with
grunt.file.expandDirs('modules/*', ...)

Pete
...from my mobile.

pkozlowski-opensource:
Fixed in 1ba6a6a and 9aab96f. Closing.

PR #185, opened 2012-09-01
Simplify custom build and make it work on win

pkozlowski-opensource:
Fixes the issue #180 and does several simplifications to the custom build started by Andy:
- code is smaller and OS-independent
- now custom build kicks-in only when arguments to the build are provided
- it is no longer necessary to specify type of a module, its name is enough: `grunt build:unique:reset:currency:highlight`

Would be great if you guys could test it on Linux / Mac. @ajoslin, would be great to have your feedback as I'm afraid that I might have simplified things too much :-)

### DIFF
--- a/grunt.js
+++ b/grunt.js
@@ -25,7 +25,7 @@ module.exports = function (grunt) {
     },
     concat: {
       build: {
-        src: ['<banner:meta.banner>', 'common/*.js', 'modules/*/*/*.js'],
+        src: ['<banner:meta.banner>'], 
         dest: '<%= builddir %>/<%= pkg.name %>.js'
       },
       ieshiv: {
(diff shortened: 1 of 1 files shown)


ajoslin:
I fixed an issue I had with custom builds.  They didn't work - it built every module, because the default concat.src still existed even if you defined a custom build.

But after that above commit, it all works for me.

pkozlowski-opensource:
@ajoslin with the 26de4c4 you've commited the build folder (we really need to do something about keeping this build in git...). I will probably drop your last commit (I know this is bad!) and re-do your changes for the common.js

ProLoser:
Can't you do --ammend if you were going to drop the commit anyway?

pkozlowski-opensource:
There is some other problem with the LESS processing, currently it is included twice in a custom build :-( 
Fixing this as well.

ajoslin:
Hopefully I can re-test this later today on my mac and we can merge this :-)

pkozlowski-opensource:
It was already merged by @petebacondarwin in 1ba6a6a3df and finished by me in 9aab96f2d3. Closing since it is finally merged.

PR #197, opened 2012-09-12
Adding TrevDev's new valid-expression validation directive

ProLoser:
WIP! Do not merge, this is for review and consideration.

Note that it may be worth while to just add support for a watch property to `ui-validate` instead of implementing this. Perhaps a second directive called `ui-validate-watch` that depends on `ui-validate` and imports the controller?

Theoretically, the 'expression' could be the functions passed to ui-validate.

### DIFF
--- a/modules/directives/validexp/validexp.js
+++ b/modules/directives/validexp/validexp.js
@@ -1,0 +1,39 @@
+/**
+ * Validation Expressions with support for watch
+ * (might get merged into ui-validate)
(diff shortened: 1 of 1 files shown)


ProLoser:
Will probably close this in lui of #198

ajoslin:
I like #198 better too; the syntax is a little less confusing (`watch` being a seperate attribute)

PR #203, opened 2012-09-19
Merging ui-date fix from pjstarifa

ProLoser:
@pjstarifa do you think you could create a new branch without the generated files committed?

Also, do you think you could add a relevant unit test for the fix you're implementing.


petebacondarwin:
Hi Peter & Dean,
The latest date directive has moved on considerably from this.  For a start
we have ditched coffeescript and also now use the datepicker to parse the
dates, which I suspect already deals with nulls. Can I suggest that you
pull the latest version from master and then work from that?  Also, I know
it is hard to grok but it is always best if you can arrange that your
changes come straight off master rather that merging other people's changes
as it makes it difficult to see what you have actually changed.  You can
use rebase for this sometimes but in this case I think a complete rewrite
is in order, sorry.
Pete

pjsvis:
Dean,

I will get latest, review and add in my changes as required, and then
commit without the build files this evening.

My change was only a few lines in the directive and the test.

Best regards,

Peter

PR #204, opened 2012-09-19
MrTheodor's dataFilter (proposed)

ProLoser:
@MrTheodor you guys gotta remember to open pull requests if you want us to merge in / review your code :)

**My notes:**
- Remove the word 'filter' from the module name since it's already in the folder (and is implied)
- In fact, drop the name 'data' altogether and use something more concise such as 'extract' or 'subset'
- This seems similar to a filter I proposed earlier called 'extract' which behaves in a similar manner to that of _.js pluck() method.

### DIFF
--- a/modules/filters/dataFilter/README.md
+++ b/modules/filters/dataFilter/README.md
@@ -5,7 +5,7 @@ Filter rules are strored in object with Django ORM like syntax.
 
 ## Example
 
-`<div ng-repeat="array|dataFilter:{id__eq:2, or:{id__eq:10}}"></div>`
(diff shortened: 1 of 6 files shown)


jkrajniak:
Yes, sure, I could not find any proper name for this filter, but 'extract' looks good. I have submitted changes.

Narretz:
Pretty cool. This line in 057b80b is incomplete, right? 
- Data filter for array of objects which is able to apply some logic data filter.

I think a good description goes a long way.

ProLoser:
@MrTheodor I'm discussing with @ajoslin about merging this, are you aware you can pass a callback function to Angular's filter ... filter?

Also, I realized that my assumption that this is similar to my old `extract` or `_.pluck()` was wrong.

PR #206, opened 2012-09-22
Travis-CI finally working!

pkozlowski-opensource:
It looks like somehow Travis-CI integration picked up, build is fine, all the tests but the map are passing: http://travis-ci.org/#!/angular-ui/angular-ui/builds/2532745

The only thing I don't understand is how the test interference I was facing just went away. I'm suspecting that it might have sth to do with a different testacular version but not sure. 

Anyway, if someone could have a look at this PR, test locally etc. @boneskull, would be great to have your opinion on this. Otherwise it should be good to merge.

After this one is merged it will be possible to remove the `travis` and `travis2` branches, yup!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,6 @@
  before_script:
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
-  - npm link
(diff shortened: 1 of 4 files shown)


boneskull:
I may not get to this until Monday but I'll see if I can tomorrow.

Issue #213, opened 2012-09-30
Date directive: element.datepicker doesn't work


frapontillo:
I'm using the latest version of anything: jQuery 1.8.2, jQueryUI 1.8.24 and Angular 1.0.2. Still, I'm having some issues with the `date` directive. The directive uses:

``` javascript
element.datepicker
```

to initialize the jQuery UI widget, but the Chrome console says "Object [[object HTMLDivElement]] has no method 'datepicker'".
If I change the directive so that it uses:

``` javascript
S(element).datepicker
```

everything works. So, is there anything wrong that I'm doing or was there some update that broke the expected behavior?

pkozlowski-opensource:
Could you try to share a jsFiddle / plunk? 
I've answered a similar question on SO recently: http://stackoverflow.com/questions/12472244/how-to-integrate-angularui-to-angularjs/12474114#12474114

There is a jsFiddle in the answer that you can use as a starting point.

frapontillo:
I have changed the jQuery plugin from 1.7.1 to 1.8.2 (which is the latest version) and I was able to replicate the behavior described in my first post: http://jsfiddle.net/frapontillo/reRPY/ .
I am submitting a pull request with the fix (compiled and tested too).

frapontillo:
I have been able to sort it out: I was referencing Angular.js before jQuery. So the element was a regular HTML element and not a jQuery one, I suppose.

petebacondarwin:
Actually element was a jqLite object - AngularJS implementation of JQuery)
used when JQuery is not found.

PR #214, opened 2012-09-30
Datepicker is now working again.

frapontillo:
I opened an [issue](https://github.com/angular-ui/angular-ui/issues/213) about the date directive not properly working under jQuery 1.8.2.
The committed changes makes it work again. There might be some additional directives which need the same modifications made here, but I haven't used any so far.


frapontillo:
It was a referencing error on my side.

PR #217, opened 2012-10-04
Support for dates of the format "1922-03-08T00:00:00.000Z" in datepicker

mchapman:
This solves a problem I was having where dates from Mongo get wrongly interpreted.  I am a bit of a beginner, so I apologise if I have done anything wrong.

### DIFF
--- a/modules/directives/date/date.js
+++ b/modules/directives/date/date.js
@@ -46,9 +46,12 @@ angular.module('ui.directives').directive('uiDate', ['ui.config', function (uiCo
           controller.$render = function () {
             var date = controller.$viewValue;
             element.datepicker("setDate", date);
-            // Update the model if we received a string
             if (angular.isString(date)) {
+              element.datepicker("setDate", new Date(date));
+              // Update the model if we received a string
               controller.$setViewValue(element.datepicker("getDate"));
+            } else {
+              element.datepicker("setDate", date);
             }
           };
         }
--- a/modules/directives/date/test/dateSpec.js
+++ b/modules/directives/date/test/dateSpec.js
@@ -109,6 +109,18 @@ describe('uiDate', function() {
         expect(element.datepicker('getDate')).toEqual(aDate);
       });
     });
+    it('should be able to use an ISOdate', function() {
+      inject(function($compile, $rootScope) {
+          var aDate, element, dateStr;
+          dateStr = "1922-03-08T00:00:00.000Z";
+          aDate = new Date(dateStr);
+          element = $compile("<div ui-date ng-model='x'></div>")($rootScope);
+          $rootScope.$apply(function() {
+              $rootScope.x = dateStr;
+          });
+          expect(element.datepicker('getDate')).toEqual(aDate);
+      });
+    });
     it('should put the date in the model', function() {
       inject(function($compile, $rootScope) {
         var aDate, element;


mchapman:
I have just noticed that all the tests pass if the 

```
controller.$setViewValue(element.datepicker("getDate"));
```

line is removed (which means that disabling the form update buttons in the way that http://docs.angularjs.org/cookbook/advancedform does it works (which at the moment it doesn't unless you are lucky with the date format).

I don't know whether that is because there is a missing test or an extraneous line of code, however.  Doesn't seem to cause me any problems so far.

petebacondarwin:
Hi Mark
Thank you for looking into this - especially providing unit tests!
 Accepting strings to date picker seems an on-going problem.  What works
for you fails for others.  It did do what you suggest a while back but then
someone wanted to provide strings in a different format to ISO, which
obviously would not work after your P.R.  The solution I came up with was
to use the date parsing inside the datepicker.  But I realise that this is
also not ideal because one may want to provide the model as a string that
is formatted differently to the view the datepicker displays.

My suggested solution is to have the date picker only accept date objects
as a model.  Then to have a new separate date-format directive, which
simply maps the model from a given formatted string to a Date object and
back again.  This works the same as how validators do. If you are going to
pass an object you just use date and ng-model.  If you want to pass a
string then you use date date-format and ng-model like so:

```
<input ui-date="datePickerConfig" ui-date-format="mm/dd/yyyy"
ng-model="myDateString" />
```

We could even set a default so that if you didn't provide a value to the
ui-date-format directive it simply formatted ISO date strings.

```
<input ui-date="datePickerConfig" ui-date-format ng-model="myISODateString"
/>
```

This should give us the best of all worlds and most power and flexibility.
 What do you think?

Pete

mchapman:
You are the one who came across people complaining, but I think that this way can be justified.  It works for unambiguous date formats (so not dd/mm/yy or mm/dd/yy) and costs nothing.  The other way costs extra code, a slightly more complex UI and gives support for the ambiguous formats (which is a mixed blessing I think).

But I know very little about the jquery ui date picker and what people like to do with it, so I may be vastly oversimplifying.

Pretty certain your suggestion is beyond my current capabilities I'm afraid.  Sorry.  I am very new to angular.

petebacondarwin:
I am talking about something like this:

``` javascript
.directive('uiDateFormat', [function() {
  var directive = {
    require:'ngModel',
    link: function(scope, element, attrs, modelCtrl) {
      if ( attrs.uiDateFormat === '' ) {
        // Default to ISO formatting
        modelCtrl.$formatters.push(function(value) {
          return value.toISOString();
        });
        modelCtrl.$parsers.push(function(value){
          return new Date(value);
        });
      } else {
        var format = attrs.uiDateFormat;
        // Use the attribute value as the format string
        modelCtrl.$formatters.push(function(value) {
          return $.datepicker.formatDate(format, value);
        });
        modelCtrl.$parsers.push(function(value){
          return $.datepicker.parseDate(format, value);
        });
      }
    }
  };
  return directive;
}]);
```

petebacondarwin:
Hi Mark,
Can you take a look at my pull request.  I think this would solve things for you, with only a small amount of extra effort, while not blocking other usages.

Example of your situation would simply be:

``` html
<input ui-date ng-model="myDate" ui-date-format />
```

mchapman:
Hi

It certainly works, and the additional requirement isn't exactly a big
deal.  I have noticed though that it refreshes 3 times - once to say
"Invalid Date" then the erroneous date conversion and finally the real
thing.

Mark

petebacondarwin:
Ok, so that needs some work!

petebacondarwin:
Can you demonstrate this flash of invalid content?
Here is my plunk: http://plnkr.co/edit/cAu2GY?p=preview
Pete

ProLoser:
Is graceful degradation setup for undefined values (the value most models
are initialized with).

Alternatively, Mark you may want to ensure you're not working with invalid
values or types when your page loads.

petebacondarwin:
Seems to degrade quite nicely.  Try different initial values in my plunk.

mchapman:
I can no longer repeat it.  Must have been something I was doing.  Sorry.

petebacondarwin:
I'll merge this then.

PR #219, opened 2012-10-04
Added ui-date-format directive to be used in conjunction with date directive.

petebacondarwin:
This is a proposed solution to the on-going problems people have with passing strings to the ui-date directive via ng-model.  The problem is how to tell the date directive how to parse strings that may have dates in a variety of formats.  See https://github.com/angular-ui/angular-ui/pull/217

The solution goes as follows:

1) The ui-date directive only accepts javascript Date objects in ng-model.  This is clean and clear.  Application developers can manipulate their scope to provide objects rather than strings.  The date directive does not conversion of strings and in fact throws an exception if one tries to pass a non-Date to it.
2) A second directive ui-date-format can be used in conjunction to ui-date to pre-parse/post-format strings from a specified date format into Date objects for the ui-date directive to consume.  This ensures that a Date object is provided to the ui-date directive and also that the scope object is always a string in the format specified.  The date-format directive would default, if not given a format string to ISO date formats.  The format strings allowed would follow what is allowed by the jquery ui datepicker.

Example usage:

``` javascript
$scope.myDate = '12 May 2012';
```

``` html
<input ui-date ng-model="myDate" ui-date-format="dd MM yy"/>
```

### DIFF
--- a/modules/directives/date/date.js
+++ b/modules/directives/date/date.js
@@ -5,7 +5,9 @@
  @param [ui-date] {object} Options to pass to $.fn.datepicker() merged onto ui.config
  */
 
-angular.module('ui.directives').directive('uiDate', ['ui.config', function (uiConfig) {
+angular.module('ui.directives')
+
+.directive('uiDate', ['ui.config', function (uiConfig) {
   'use strict';
   var options;
   options = {};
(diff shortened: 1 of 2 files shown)


petebacondarwin:
Does anyone have a problem  with this change to the date directive?
It is a breaking change for anyone expecting to pass a string to it.
If ok then I'll update the read and demo then merge into master.
Pete

petebacondarwin:
I have merged this branch (via rebase) onto the master.

Time for a version release?

ProLoser:
I've got no problem with you bumping the version. Assuming I understand this fix correctly, looks like I may need it. Does this simply allow me to use a string as my value on the model instead of the object?

petebacondarwin:
Yes.

Pete
...from my mobile.

PR #225, opened 2012-10-09
Keep ui-date input field in sync with date picker for manual input

hvisser:
When entering a date manually in an input field the date picker isn't always updated and will put the previously selected date in the model. Likewise, the input field itself isn't updated reflecting the date which is in the model. This pull request fixes that by setting the date on the date picker in the change event and if the date really changed, calling updateModel to get the Angular model updated.

### DIFF
--- a/modules/directives/date/date.js
+++ b/modules/directives/date/date.js
@@ -40,7 +40,13 @@ angular.module('ui.directives').directive('uiDate', ['ui.config', function (uiCo
             opts.onSelect = updateModel;
           }
           // In case the user changes the text directly in the input box
-          element.bind('change', updateModel);
+          element.bind('change', function() {
+              var date = element.datepicker("getDate");
+              element.datepicker("setDate", element.val());
+              if (!angular.equals(date, element.datepicker("getDate"))) {
+                  updateModel();
+              }
+          });
 
           // Update the date picker when the model changes
           controller.$render = function () {
--- a/modules/directives/date/test/dateSpec.js
+++ b/modules/directives/date/test/dateSpec.js
@@ -58,6 +58,21 @@ describe('uiDate', function() {
     });
   });
 
+  it('should update the input field correctly on a manual update', function() {
+      inject(function($compile, $rootScope) {
+          var aDateString, element;
+          aDateString = '2012-08-17';
+          element = $compile('<input ui-date="{dateFormat: \'yy-mm-dd\'}" ng-model="x"/>')($rootScope);
+          $rootScope.$apply(function() {
+              $rootScope.x = aDateString;
+          });
+          element.val('2012-8-01');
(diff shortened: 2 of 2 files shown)


petebacondarwin:
@hvisser - Thank you for identifying this problem and suggesting a fix.  I wasn't exactly sure what the problem is but I think you meant that if you enter text into the date input box, then calendar is not updating correctly?  I couldn't reproduce this problem.  But I did notice that if you enter a date that is not quite the correct format, e.g. 11/8/2012 rather than 11/08/2012, if the format is set to dd/mm/yy, then the date picker updates itself but it doesn't reformat the input box accordingly.  Is this your problem?

Anyway, I have an outstanding branch on this directive - https://github.com/angular-ui/angular-ui/pull/219, which moves things forward regarding date strings in the model.  I slightly modified and applied your patch and I believe it is working now.  Can you take a look?

Cheers,
Pete

hvisser:
Hi Pete, the input not in sync thing definitely is what I'm trying to solve here. Please disregard the other one (date not set in the model), I think that was just me being confused looking at the same problem before.

In your fix you call the "setDate" in updateModel, which I did before, however I was under the impression that that triggered multiple passes of updateModel so that's why I used the equal check and a separate function.

Btw, I don't care much about the model as a string being parsed to a date and I actually don't think the directive should do that. It really gets confusing on what should be in the model imho. I'd expect my model always to be a date object, since that what's returned too by the date picker. Put a date in, get a date back, simple ;)
There might be a use case for converting a bound string to a date, but maybe that kind of conversion is better off in a controller or a separate directive?

hvisser:
@petebacondarwin Oh, actually now that I look at your date branch, this is exactly what you're doing, separating the two! Excellent ;)

petebacondarwin:
Yes, I agree that the date picker should only work with Date objects but
people do have a tendency to pass in strings - like you did in your
proposed unit test :-)
This new version adds an extra date-format directive that wraps around the
model value for the datepicker to ensure that the directive only ever gets
a Date object but that users can have formatted strings in their model.
The ngModelController already checks to see if the model has changed before
triggering its own update and since we are now only using Date objects (and
not a confusing mix of Dates and strings) it is able to see that the model
has not changed and so stops it there.
Pete

hvisser:
Well the "editing the date in the input field" case I don't regard as binding in this case. I do think the date directive should take care of that from a component perspective and make sure that whatever is filled in that field is either something valid or nothing at all. I think that is what I'm testing right? :)

What I object to in the current implementation that the property you are binding to in ng-model can be a string or a date, that is really confusing, although not something that is in my way. If you are trying to fix that, awesome, but I think the ui-date directive itself should do the right thing by default when a user types in the date in the input field.

PR #229, opened 2012-10-15
Allow select2 to use filters in an *ng-repeat statement

jaymes-bearden:
Previous to this commit, the split statement on the *ng-repeat did not split on the filtering character ( | ).

This small commit properly filters out only the data that should be watched even when filters are applied.


ProLoser:
Really need to add unit tests to this directive. @jaymes-bearden I'm going to merge in your code but if you think you could add any unit tests too that would be much appreciated.

PR #232, opened 2012-10-17
select2 "disabled" attribute improved

nidi3:
The disabled attribute is only taken into account if it changed after initialization, so
<input ui-select2 disabled="true"> 
does NOT work.

The disabled attribute is interpreted as string, so every value apart from null and empty string is true, so
<input ui-select2 disabled="{{booleanFlag}}">
is always disabled.

This change fixes both issues.

### DIFF
--- a/modules/directives/select2/select2.js
+++ b/modules/directives/select2/select2.js
@@ -87,7 +87,7 @@ angular.module('ui.directives').directive('uiSelect2', ['ui.config', '$http', fu
         }
 
         attrs.$observe('disabled', function (value) {
-          elm.select2(value && 'disable' || 'enable');
+          setDisabled(elm, value);
         });
 
         // Set initial value since Angular doesn't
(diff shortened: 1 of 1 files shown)


ProLoser:
Refer to this demo to get an idea of when 'disabled' actually means 'disabled'

http://jsfiddle.net/Y7Z3G/

If you view the non ng-prefixed versions, you will see that every variety (including true, false, empty string or no value) are all considered disabled. If you want to use an expression you SHOULD be using ng-disabled (for which the original code I had in place should work fine).

Now if I'm wrong feel free to add any notes, I also confirmed that the initial value of disabled is checked. I do not need to do this myself because select2 checks to see if the element is disabled upon initialization and $observe will fire otherwise.

it MAY be worthwhile to $observe ngDisabled instead of simply disabled as ngDisabled is liable to change value while disabled may not, however since the current setup works, it covers more potential edge cases.

### tl;dr

You're supposed to use `ng-disabled`

PR #240, opened 2012-10-24
Attempt to fix running tests on mac

ProLoser:
This should be a quick/easy merge if someone can confirm this is even necessary? I am unable to run tests without this fix.

### DIFF
--- a/grunt.js
+++ b/grunt.js
@@ -108,7 +108,7 @@ module.exports = function (grunt) {
   grunt.registerTask('test', 'run tests (make sure server task is run first)', function () {
     var done = this.async();
     grunt.utils.spawn({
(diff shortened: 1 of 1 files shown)


pkozlowski-opensource:
@ProLoser Dean, it is about the testacular version you are using, not about Mac. Just install globally the testacular version referenced in package.json: `~0.2.0`

pkozlowski-opensource:
@ProLoser Dean, as I have mentioned, you just need to update the testacular version.

Going to close this PR as merging it would break Travis-CI and builds of all people already using the latest stable testacular (in fact we should bump up testacular version to 0.4.x since this is the latest stable).

Issue #181, opened 2012-08-28
uiModal should take care of autofocus


prasad83:
``` javascript
angular.module('ui.directives')
.directive('uiModal', ['$timeout', function($timeout) {
  return {
    // ...
    link: function(scope, elm, attrs, model) {
      // ...
      elm.on('show.ui', function() {
        $timeout(function() {
          model.$setViewValue(true);
            // FIX: Trigger focus for autofocus elements
            elm.find('[autofocus]').focus();
        });
      });
      // ...
    }
  };
}]);
```

ProLoser:
1: I don't think that code will work. 2: I don't know that it should be in
the directive. Seems sort of outside scope. Can't you use the faded event
or something to focus it?

On Tue, Aug 28, 2012 at 12:02 AM, prasad83 notifications@github.com wrote:

> angular.module('ui.directives').directive('uiModal', ['$timeout', function($timeout) {
>   return {
>     // ...
>     link: function(scope, elm, attrs, model) {
>       // ...
>       elm.on('show.ui', function() {
>         $timeout(function() {
>           model.$setViewValue(true);
>             // FIX: Trigger focus for autofocus elements
>             elm.find('[autofocus]').focus();
>         });
>       });
>       // ...
>     }
>   };}]);
> 
>  —
> Reply to this email directly or view it on GitHubhttps://github.com/angular-ui/angular-ui/issues/181.

## 

Dean J Sofer
DeanSofer@gmail.com

BS Computer Information Systems
California State Polytechnic University, Pomona
Telephone: 714-900-2254
Website: www.DeanSofer.com
Github: github.com/ProLoser

prasad83:
I have been using ui-modal bound to ng-model, in this approach I expect the autofocus to work when the modal shows up:

``` html
<div ui-modal ng-model="showEditModalForm">
    <form class="form-stacked" ng-submit="createRecord()">
        <input type="email" ng-model="editDeal.fromaddr" required="required" class="span5" autofocus="autofocus">
    </form>
</div>
```

Following enhanced fix worked for me:

``` javascript
angular.module('ui.directives')
.directive('uiModal', ['$timeout', function($timeout) {
  return {
    // ...
    link: function(scope, elm, attrs, model) {
      // FIX to avoid repeated autofocus
      var afocused = false;
      // ...
      elm.on('show.ui', function() {
        $timeout(function() {
          model.$setViewValue(true);
[...]

PR #244, opened 2012-10-25
modal: add support for autofocus

slav:
automatically set focus on element with [autofocus] attribute when modal is shown.


ProLoser:
I think we are completely relocating modal to angular-ui/bootstrap

You may want to consider opening this PR over there. Sorry for the hassle :)

slav:
It's different there. I'll wait until there's an official release with some docs to add support for autofocus and send pull request.

ProLoser:
This should address #181

PR #253, opened 2012-11-01
Replaced trim() method in select2 directive

escalant3:
Trim is not available in IE8 and below. This causes the select2 to fail
when <option ng-repeat> is used.

This commit replaces the trim method with a standard unobtrusive regex.


ProLoser:
Might as well use jQuery.trim() since it's already a dependency.

ProLoser:
5279043da2885d7bdeb487a3569ec9dea83e6217

PR #256, opened 2012-11-04
Bootstrap datepicker

slav:
Created directive to show bootstrap datepicker (https://github.com/eternicode/bootstrap-datepicker) and bind it to AngularJS via attributes.

Attributes are used to customize how datepicker is shown.

There're no tests and code isn't fully tested. Don't have time to work on that. It works for me. If I encounter any issue will fix and push update.

### DIFF
--- a/modules/directives/datepicker/datepicker.js
+++ b/modules/directives/datepicker/datepicker.js
@@ -1,0 +1,63 @@
+﻿'use strict';
+
+angular.module( 'ui.directives', [] )
+.directive( 'uiDatepicker', ['ui.config', function( uiConfig ) {
+ 	var options;
+ 	options = {};
+ 	if( angular.isObject( uiConfig.datepicker ) ) {
+ 		angular.extend( options, uiConfig.datepicker );
+ 	}
+ 	return {
+ 		require: '?ngModel',
+ 		link: function( $scope, element, attrs, controller ) {
+ 			var getOptions = function() {
+ 				var options = {};
+
+ 				if( attrs.uiFormat ) options.format = $scope.$eval( attrs.uiFormat );
+ 				if( attrs.uiWeekStart ) options.weekStart = $scope.$eval( attrs.uiWeekStart );
+ 				if( attrs.uiStartDate ) options.startDate = $scope.$eval( attrs.uiStartDate );
+ 				if( attrs.uiEndDate ) options.endDate = $scope.$eval( attrs.uiEndDate );
+ 				if( attrs.uiAutoclose ) options.autoclose = $scope.$eval( attrs.uiAutoclose );
+ 				if( attrs.uiStartView ) options.startView = $scope.$eval( attrs.uiStartView );
+ 				if( attrs.uiTodayBtn ) options.todayBtn = $scope.$eval( attrs.uiTodayBtn );
+ 				if( attrs.uiTodayHighlight ) options.todayHighlight = $scope.$eval( attrs.uiTodayHighlight );
+ 				if( attrs.uiKeyboardNavigation ) options.keyboardNavigation = $scope.$eval( attrs.uiKeyboardNavigation );
+ 				if( attrs.uiLanguage ) options.format = $scope.$eval( attrs.uiLanguage );
+
+ 				options = angular.extend( options, uiConfig.datepicker, $scope.$eval( attrs.datepicker ) );
+
+ 				return options;
+ 			};
+
+ 			var opts = getOptions();
+ 			if( controller ) {
+ 				var updateModel = function( ev ) {
+ 					return $scope.$apply( function() {
+ 						var date = ev.date;
+ 						controller.$setViewValue( date );
+ 					} );
+ 				};
+
+ 				controller.$render = function() {
+ 					var date = controller.$viewValue;
+ 					if( date )
+ 						element.datepicker( 'setDate', date );
+ 				};
+ 				if( attrs.uiStartDate ) {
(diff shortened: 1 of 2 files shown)


ProLoser:
You should open this against the bootstrap project instead: https://github.com/angular-ui/bootstrap

slav:
Will do, but unsure what is the future of it and what to expect. Any real commits are more than 2 weeks old and there're no docs, demos or releases. I'm using AngularJS UI, not using UI bootstrap (since no releases).

ProLoser:
I'll let @ajoslin chime in on that one as he and @pkozlowski-opensource are sort of spearheading that project. I do believe it's just as active as AngularUI. You could have said the same thing (in reverse) about AngularUI vs Bootstrap however I finally had the first break from work for months and have been trying to catch up on things.

There are a few people interested in bootstrap so I don't think it's any less active. There are also unit tests on Google's very own CI servers while we still have yet to get AngularUI's tests on there.

slav:
Well, without build, docs, or demo it has a feeling of "alpha". I'll follow up with them, perhaps something can be done to improve the situation.

ProLoser:
Understandable. Keep in mind if you're contributing to the project (and
lend a hand with those things) we can add you to the core team!

On Sat, Nov 3, 2012 at 8:36 PM, slav notifications@github.com wrote:

> Well, without build, docs, or demo it has a feeling of "alpha". I'll
> follow up with them, perhaps something can be done to improve the situation.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/angular-ui/angular-ui/pull/256#issuecomment-10047135.

## 

Dean J Sofer
DeanSofer@gmail.com

Telephone: 714-900-2254
Website: www.DeanSofer.com
Github: github.com/ProLoser

slav:
Well, I'm extremely busy at work (usual startup craziness). So, I quickly develop things I need and move on. I'm happy to share what I got. I usually try to make sure directive or whatever is simple to understand and use. But I really don't have time to write tests or polish it to the level needed to make it part of the release without additional work.

pkozlowski-opensource:
@slav thnx for taking time to open this pull requests. Unfortunately we can't accept it in this form for 2 main reasons:
- the goal of the boostrap repository is to provide "native" AngularJS directives, that is - ones that don't have any dependency on the external JavaScript libraries. Your directive has a direct dependency on the external JavaScript code;
- We need code to be tested properly. We've got a CI server setup at Google that runs all the tests on different browsers. This is our safety net to ensure that we are pushing bug-free code and that we don't introduce regressions when the code evolves.

I'm glad to hear that your solution works for you but it is true that writing for the opens source community requires some additional effort. 

Going to close this PR for now, fell free to reopen it when you've got time to write tests for your code. I'm going to create contributors guide to make our policy clear.

slav:
Well, even if I write tests, it still won't fit in since it relies on external (not native BS) library. So in other words, this directive has no place in this project, even if it would be properly polished and tested?

slav:
Thinking about it, perhaps create a separate branch specifically for unfinished and / or with external dependencies directives?

pkozlowski-opensource:
@slav, we've got already a date-picker directive in the angular-ui repository (one based on jQuery). Maybe you should compare the 2 and propose replacing the jQuery one with the boostrap one it this has advantages?

Regarding the bootstrap repo - yes, you are right, it wouldn't fit since it has direct dependencies on the external JS.

slav:
@pkozlowski-opensource I don't think it replaces jQuery datepicker. Some people use jQuery UI, and they would use jQuery based plugin. Some people use bootstrap, and, like me, they would prefer bootstrap based datepicker. It all depends on context.

About external dependency, what if external plugin is fully included (js and css)? Or you want to avoid this situation as well?

scottromack:
+1 for bootstrap datepicker
Here is another one that has a timepicker available from the same author.

http://www.eyecon.ro/bootstrap-datepicker/
Not saying this one is better but it follows the bootstrap look...

scottromack:
One advantage is avoiding the style fragmentation of jquery UI.

pkozlowski-opensource:
@ShaggyDude we've got a separate repository (https://github.com/angular-ui/bootstrap) for the bootstrap's specific directives. Demo here: http://angular-ui.github.com/bootstrap/

But in this repository we've taken a decision to not include jQuery nor 3rd party JavaScript as a dependency. So if the mentioned plugin can be made to work without jQuery and Bootstrap's JavaScript I would be all for adding it! A pull request would be highly appreciated.

leesei:
It's a pity that this pull got turned down.
The current Tweeter Bootstrap (and hence Angular UI Bootstrap) lacks a official datepicker.
UI Bootstrap decided not to include 3rd party dependencies, so we have to re-implement bootstrap-datepicker in jqlite if we have to put it in the repo.

Angular UI is more lenient on 3rd party dependencies, but have chosen to favor on jQuery UI.
Now there is a Angular UI Bootstrap project, all Bootstrap plugins will be forwarded there (and rejected if not re-implemented).
This is an unfortunate scenario.

I'm using jQuery + Angular UI + Bootstrap +Angular UI Bootstrap (which I think is a common combo) in my project.
I've manually added @slav's `datepicker.js` to my project and it worked quite well (yes, there're some bugs, but we can fix it).

Knowing and respecting the design goal of UI Bootstrap, here's a little suggestion:
Can Angular UI/UI Bootstrap team help and maintain a repo of Bootstrap plugin directives so they can be easily found?
I've Google quite a bit to have stumbled upon this thread, it would be great if quality plugin directives can be found more easily.

PR #264, opened 2012-11-15
Delayed setting select2 values

nkatsaros:
When a select2 box is behind a ui-if directive the values won't get set once it appears. 

setTimeout fixes the problem.


ProLoser:
Sorry @nkatsaros, as per our discussion can't really merge this as it's fixing a symptom rather than a problem. `ui-if` will continue to create unusual behavior across many third-party plugins.

PR #275, opened 2012-11-22
Jq: Do not copy element options to the defaults

pracj3am:
When using multiple Jq directives, elements options (from `ui-options` attribute) are copied to the global default options and then they are applied to all latter directive instances.


ajoslin:
Good fix.  Could you add a test case for it?

Something like:

``` js
describe('should not copy options to global', function() {
  $.fn.joe = function(opts) {}
  $.fn.susan = function(opts) {
    it('should not copy opts.pizza', function() {
      expect(opts.pizza).toBeFalsy();
    });
  }
  $compile('<div ui-jq="joe" ui-options="{pizza:true}"></div>')($scope);
  $compile('<div ui-jq="susan" ui-options="{}"></div>')($scope);
});
```

pracj3am:
Sure

ProLoser:
Try to update testacular

ProLoser:
Heh thanks for this fix, sorry for the tardy merge. This actually screwed me up on a presentation I gave.

PR #279, opened 2012-11-28
event fix

jbu:
I had an issue in which inserting a row into an ng-repeat table when elements of the table had focus handlers bound caused a conflicting $scope.apply. This is one possible fix. 
Either some of these applies need to be rebased by timeout, or I'm doing something wrong. Probably the latter, but this works for me.

### DIFF
--- a/modules/directives/event/event.js
+++ b/modules/directives/event/event.js
@@ -8,8 +8,8 @@
  *
  * @param ui-event {string|object literal} The event to bind to as a string or a hash of events with their callbacks
  */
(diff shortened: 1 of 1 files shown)


ProLoser:
Actually the proper way to fix this would be to check if you're already in an `$apply` before calling one.

Read up here on how to do this: http://deansofer.com/posts/view/14/AngularJs-Tips-and-Tricks-UPDATED#debugging

```
if (!$scope.$$phase) $scope.$apply()
```

jbu:
sounds more reasonable. I thought i was doing something wrong. i'll give it
a go.

PR #283, opened 2012-11-30
adds a ellipsize filter

alexfigueiredo:
this is borrowed from codeigniter's helpers. I found it useful, maybe it
can help others.

### DIFF
--- a/modules/filters/ellipsize/ellipsize.js
+++ b/modules/filters/ellipsize/ellipsize.js
@@ -1,0 +1,46 @@
+﻿/**
+	 * Ellipsize String
+	 *
+	 * This function will strip tags from a string, split it at its max_length and ellipsize
+	 *
+	 * @param	string	string to ellipsize
+	 * @param	int	max length of string
+	 * @param	mixed	int (1|0) or float, .5, .2, etc for position to split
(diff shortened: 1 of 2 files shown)


alexfigueiredo:
I'm sorry, forgot to jslint my code. Closing to do it again.

ProLoser:
Are you aware you can do this entirely in css?

damrbaby:
Actually you cannot do this purely in CSS if the text spans multiple lines... see http://stackoverflow.com/questions/5269713/css-ellipsis-on-second-line

alexfigueiredo:
Also, css allow you to add the ellipsize only to the end of the string.

This filter allow you to do this:

"This is a very lon...string!"

ProLoser:
@alexxfsilva I'm going to close this here, can you instead relocate this to @angular-ui/ui-utils ?

Also, I would try to shave the length of the filter name as much as possible (even `ellipsis`), but that's just me.

You also need tests for all the variations of ways you can call this filter.

And finally, I would like to see it pass-thru gracefully when the length property is falsy, this way someone can easily toggle the ellipsis on and off instead of having 2 DOM elements they must hide and show.

PR #288, opened 2012-12-04
New filter captureHighlight

garethdjames:
Added a new filter for highlighting based on a capturing regular expression


ProLoser:
Is there a reason this could not supplement the existing highlight filter rather than making a completely new one?

garethdjames:
I agree with your suggestion to supplement the existing highlight filter. I will do a new pull request tomorrow. Probably best to drop this request as there will not be a new filter

PR #291, opened 2012-12-05
Make working sortable directive

lstrzelecki:
Fix Issue#255 Problem with sortable directive

### DIFF
--- a/modules/directives/sortable/sortable.coffee
+++ b/modules/directives/sortable/sortable.coffee
@@ -10,7 +10,7 @@ angular.module('ui.directives').directive 'uiSortable', ['ui.config', (uiConfig)
 
   require: '?ngModel'
   link: (scope, element, attrs, ngModel) ->
-    opts = angular.extend({}, options, scope.$eval(attrs.uiOptions))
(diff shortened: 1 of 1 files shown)


ProLoser:
Although the directive already worked before, I agree that the options should just be passed directly.

lukeholder:
Would love to see a nested sortable also: http://mjsarfatti.com/sandbox/nestedSortable/

ProLoser:
@lukeholder consider starting a new issue instead of piggybacking onto the end of this now closed and old thread.

There's also a ui-sortable repo you can add it to.

raffus:
Hello @ProLoser, I'm having a hard time getting angularjs sortable to work.
I get this weird behavior that everytime I drop a draggable item it duplicates the item with blank content. Also, sortable items just work with this duplicated items, not with original. I mean, I drop the original draggable item and it will get duplicated. The duplicated item can be sortable but it is useless as it is right now.

Looking at other posts, I found something related to the 'connectWith' directive, but even without this it wont work.

I'm using connected lists.

Any help would be very appreciated.
BR,
Rafael

PR #292, opened 2012-12-05
Highlight Filter Capture Regex

garethdjames:
Added a parameter to support highlighting based on a capturing regular expression

### DIFF
--- a/modules/filters/highlight/highlight.js
+++ b/modules/filters/highlight/highlight.js
@@ -3,21 +3,31 @@
  * @param text {string} haystack to search through
  * @param search {string} needle to search for
  * @param [caseSensitive] {boolean} optional boolean to use case-sensitive searching
- * @param [capture] {boolean} optional boolean to use a regular expression capture for highlighting
  */
 angular.module('ui.filters').filter('highlight', function () {
     return function (text, search, caseSensitive, capture) {
         if (search || angular.isNumber(search)) {
-            text = text.toString();
-            search = search.toString();
-            try {
-                if(capture && (new RegExp(search, caseSensitive && 'g' || 'gi').exec(text).length == 1)) {
-                    return text;
+            var start = '<span class="ui-match">',
+                end = '</span>',
+                result,
+                reg,
+                match;
+            try{
+                text = text.toString();
+                search = search.toString();
+                if (caseSensitive && !capture) {
+                    result = text.split(search).join(start + search + end);
+                } else {
+                    reg = new RegExp(search, caseSensitive && 'g' || 'gi');
+                    match = reg.exec(text);
+                    result = text.replace(reg, start + (capture && '$1' || '$&') + end);
+                    if(capture && match[0] != match[1]) {
+                        result = text.substring(0, reg.lastIndex - match[1].length) + result;
+                    }
                 }
-                return text.replace(new RegExp(search, caseSensitive && 'g' || 'gi'), '<span class="ui-match">' + (capture && '$1'|| '$&') + '</span>');
-            }
-            catch (Exception) {
-                return text.toString();
+                return result;
+            } catch (Exception) {
+                return text;
             }
         } else {
             return text;
--- a/modules/filters/highlight/highlight.js
+++ b/modules/filters/highlight/highlight.js
@@ -7,41 +7,17 @@
  */
 angular.module('ui.filters').filter('highlight', function () {
     return function (text, search, caseSensitive, capture) {
-        if(!capture) {
-            if (search || angular.isNumber(search)) {
-                text = text.toString();
-                search = search.toString();
-                if (caseSensitive) {
-                    return text.split(search).join('<span class="ui-match">' + search + '</span>');
-                } else {
-                    return text.replace(new RegExp(search, 'gi'), '<span class="ui-match">$&</span>');
-                }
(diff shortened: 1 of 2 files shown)


ProLoser:
Okay, why the hell did Travis only run 1 test?

modules/filters/highlight/highlight.js
```diff
@@ -3,19 +3,45 @@
...
+            } else {
+                return text;
+            }
+        } else {
+            try {
+                if (search || angular.isNumber(search)) {
+                    text = text.toString();
+                    search = search.toString();
+                    var match;
+                    if (caseSensitive) {
```
ProLoser: Perhaps it's more efficient to do:

```
var scope = 'g';
if (!caseSensitive)
  scope += 'i';
// or
var scope = caseSensitive && 'g' || 'gi';
```

Rather than duplicate a more complex line of logic.

modules/filters/highlight/highlight.js
```diff
@@ -3,19 +3,45 @@
...
+                        match = new RegExp(search, 'g').exec(text);
+                    } else {
+                        match = new RegExp(search, 'gi').exec(text);
+                    }
+                    var firstPart = text.substr(0, text.indexOf(match[1]));
+                    var secondPart = '<span class="ui-match">' + match[1] + '</span>';
+                    var lastPart = text.substr(text.indexOf(match[1]) + match[1].length, text.length);
```
ProLoser: Is it not more efficient to use replace() with 2 parameters instead of this?

modules/filters/highlight/highlight.js
```diff
@@ -3,19 +3,45 @@
...
+            if (search || angular.isNumber(search)) {
+                text = text.toString();
+                search = search.toString();
+                if (caseSensitive) {
+                    return text.split(search).join('<span class="ui-match">' + search + '</span>');
+                } else {
+                    return text.replace(new RegExp(search, 'gi'), '<span class="ui-match">$&</span>');
+                }
+            } else {
+                return text;
```
ProLoser, line 30: **OCD:** can cut down the line count by doing:

```
text = text.split(search).join(etc)
...
return text
```

Rather than putting else-cases everywhere (in case we add more cases later).

NARKOZ:
@ProLoser you can go to Travis build page and re-run the build under the gear icon.

garethdjames:
I have updated the code to reduce the LOC and support the new capture option.
The new code always uses a regex but reduces the LOC and complexity

garethdjames:
Don't merge just yet, there is a change I want to make

garethdjames:
The code before whilst passing all tests was failing in production for a capture highlight at the beginning of the text. I have added new tests for this case and updated the code.

The new code results in 2 regex searches for capture highlight, this could be reduced to 1 regex but would increase the LOC with if/else

ProLoser:
@joseym Got another one for you. I just haven't had the time to review the source code and to figure out the new API

joseym:
Okay, I'll have a look over the weekend.

joseym:
Looks good to me. @garethdjames - from what I gather the `capture` option forces the match to be a regex, yes?

I played with it in a fiddle: http://jsfiddle.net/joseym/T5uGQ/

I noticed that `([0-9]{3}-[0-9]{3}-[0-9]{4})` will highlight the results regardless of whether or not `capture` is `true`, however `[0-9]{3}-[0-9]{3}-[0-9]{4}` will only highlight the results if `capture` is `false`.

I kindof wonder what the point is ... if the regular match accepts regular expressions why have an option to force the match to be a regular expression?

ProLoser:
@joseym actually i forgot about that...

Maybe we shouldn't merge this in? Just add some more tests / demos? Perhaps it would be better to simply tweak the case sensitivity option to be like `caseInsensitiveOrREGEX` etc

ProLoser:
I'm going to defer this for now to v0.4.1 since it doesn't really introduce any real new functionality and needs to be discussed a bit further.

joseym:
Yeah, with regex ability already in the filter I'm not sure what the capture param gives us.

garethdjames:
The purpose of the parameter was to enable a case where you wanted to enter a regex that would include a capture, i.e. to capture part (or all) of the data.  

My use case involves a user writing a regex that will filter (using a () capture) part of the string (in this case a URL). I wanted to show the captured part as highlighted.

e.g.

given the string http://www.somesite.com/index.html?a=b&c=d

and the regex http://(.*)\?

the highlight would look like this

http://www.somesite.com/index.html?a=b&c=d  

(I'm using 'bold' as the highlight in this email).

If you think this is something worth adding then I can rework it, otherwise I will just add a filter to my own project

Cheers

Gareth

joseym:
Ahh, I see now. I actually think that is rather good, useful, addition now that I understand the point :smile:. I updated the fiddle as well.

joseym:
I noticed in the fiddle that after the capture is applied the url string is somewhat scattered in the paragraph and some characters that once existed are lost altogether (the ? kicking off the url query is missing after the filter fires)

garethdjames:
Yes needs a bit more work to get it right

Will update later

ProLoser:
If this is still a concern can you try adding it to @angular-ui/ui-utils

PR #304, opened 2012-12-12
fix bug 'has no method fullCalendar'

jrschumacher:
added jQuery wrapper

### DIFF
--- a/modules/directives/calendar/calendar.js
+++ b/modules/directives/calendar/calendar.js
@@ -48,7 +48,7 @@ angular.module('ui.directives').directive('uiCalendar',['ui.config', '$parse', f
               //extend the options to suite the custom directive.
               angular.extend(options, uiConfig.uiCalendar, expression);
               //call fullCalendar from an empty html tag, to keep angular happy.
-              elm.html('').fullCalendar(options);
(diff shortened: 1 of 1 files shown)


ProLoser:
Are you sure this isn't a bug with AngularJS's jqlite not returning the original object when passing a string to `.html()`? The code looks fine to me, and I confirmed it.

jrschumacher:
oh I guess it is then, I wasn't aware of that functionality.

// edit, I see what you are saying

ProLoser:
I haven't tested if this IS in fact a bug with AngularJS, I just confirmed that this works properly when vanilla jQuery is loaded. Feel free to add any notes you collect or open a bug on AngularJS?

joshkurz:
Here is a link to most updated documentation for the calendar to date. I will start working on the demo for the angular-ui website. http://www.dillingermediaonline.com/atlanta-web-developer/angularjs-calendar-2261/

jrschumacher:
@ProLoser Yes that is the issue. Apparently I am having this problem popup all over. What version of Angular did you test with?

ProLoser:
I didn't test to confirm that Angular's version is behaving properly, I only confirmed that the jQuery behavior works with the previously existing code.

PR #307, opened 2012-12-14
This fixes a missing scope.$apply() in the ui-date onSelect event

djsmith42:
Prior to this commit, callers' onSelect handler would need to use its
own $scope.$apply() for changes to $scope to take effect as a result of
an onSelect event.

### DIFF
--- a/modules/directives/date/date.js
+++ b/modules/directives/date/date.js
@@ -38,7 +38,7 @@ angular.module('ui.directives')
             opts.onSelect = function (value, picker) {
               updateModel();
               scope.$apply(function() {
-                  userHandler(value, picker);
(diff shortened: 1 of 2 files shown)


petebacondarwin:
Nice one @djsmith42!  Actually looking at it, we don't need to return the value returned from the user function.  The data picker doesn't care.  So we can ditch the `ret` variable bit.  Fancy changing this and then adding a unit test to the date directive?  Then I'll push it in.
Pete

djsmith42:
I'll clean this up and re-submit

petebacondarwin:
Thanks! Look forward to seeing the new PR!

PR #308, opened 2012-12-14
TDD fix for missing scope.$apply() in ui-date onSelect

djsmith42:
Previously, the onSelect caller would have to do its own $scope.$apply(). Now we do it for them to be consistent with the rest of Angular.

### DIFF
--- a/modules/directives/date/date.js
+++ b/modules/directives/date/date.js
@@ -37,7 +37,9 @@ angular.module('ui.directives')
             var userHandler = opts.onSelect;
             opts.onSelect = function (value, picker) {
               updateModel();
-              return userHandler(value, picker);
+              scope.$apply(function() {
+                userHandler(value, picker);
+              });
             };
           } else {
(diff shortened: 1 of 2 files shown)


petebacondarwin:
Great!  I'll merge this evening.

pkozlowski-opensource:
@djsmith42 awesome PR, very clean! There is a fix but there are also tests, good commit messages and 2 separate commits for a test and the actual fix. Very, vary much appreciated! :+1:

ProLoser:
Is there a chance `onSelect` might be triggered by an $apply() already in progress? And should updateModel() be wrapped in the apply too?

petebacondarwin:
updateModel has a $apply in it already.

petebacondarwin:
The normal process for onSelect is from a user clicking on a date.  It
would have to be some weird simulation thing going on to cause it to be
inside AngularJS already.  One could say the same about any jQuery event.

PR #309, opened 2012-12-14
Select2: Use the val method for array $modelValue

ggoodman:
The check, `angular.isObject(controller.$modelValue)` will return
`true` if `controller.$modelValue` is an array.

Select2 insists on using the `val` method for arrays.

### DIFF
--- a/modules/directives/select2/select2.js
+++ b/modules/directives/select2/select2.js
@@ -49,6 +49,8 @@ angular.module('ui.directives').directive('uiSelect2', ['ui.config', '$http', fu
             } else {
               if (isMultiple && !controller.$modelValue) {
                 elm.select2('data', []);
+              } else if (angular.isArray(controller.$modelValue)) {
+                elm.select2('val', controller.$modelValue);
               } else if (angular.isObject(controller.$modelValue)) {
                 elm.select2('data', controller.$modelValue);
               } else {


ProLoser:
This came up with me a little. The question is are you doing an array of strings or an array of objects? What about a string with multiple items and a separator?

ProLoser:
Pinging @ggoodman because having open PR's makes me a sad panda :(

ProLoser:
I think this may be better:

``` js
controller.$render = function () {
  if (isSelect) {
    elm.select2('val', controller.$modelValue);
  } else {
    if (isMultiple) {
       if (!controller.$modelValue) {
        elm.select2('data', []);
      } else if (angular.isArray(controller.$modelValue)) {
        elm.select2('data', controller.$modelValue);
      } else {
        elm.select2('val', controller.$modelValue);
      }
    } else {
      if (angular.isObject(controller.$modelValue)) {
        elm.select2('data', controller.$modelValue);
      } else {
        elm.select2('val', controller.$modelValue);
      }
    }
  }
};
```

The only thing is I don't remember if 'val' takes an array or a comma-separated string for multiple.

See this demo: http://plnkr.co/edit/Ni4QubpSRpfYcLgm4QXd (doesn't have multiple)

Issue #312, opened 2012-12-16
CodeMirror support broken with version 3.0


prasad83:
onChange handler is not getting triggered with codemirror 3.0 which breaks ui-codemirror model update.

The fix below worked:

``` js
codemirror = CodeMirror.fromTextArea(elm[0], options);
// Bug fix - for codemirror 3
if (typeof codemirror.on == 'function') {
    codemirror.on('change', onChangeHandler);
}
```

ProLoser:
So definitely seems that there was a breaking API change with CodeMirror.

My question is should we reflect this by breaking support for v2 with this change?

ProLoser:
Nevermind looks like your proposed fix is non-destructive. Have you thought of opening a PullRequest with the fix in the future?

douglasduteil:
Yep I had the same problem. Here is how I solved it...
Demo : http://jsbin.com/aqecex/2/
Code : https://gist.github.com/2a00bea5db6293b6f4b3

ProLoser:
@douglasduteil want to open a PR with the fix?

douglasduteil:
Sure, why not :)
But it's not really fixed...
It seems like after the connection is done the compiler just don't care about the rest... 

Demo : http://jsbin.com/aqecex/5/

douglasduteil:
K !
This only occurs in the parent node !
Demo : http://jsbin.com/aqecex/6/

PR #314, opened 2012-12-18
CodeMirror 3.0 support

douglasduteil:
### Update the CodeMirror version supported from 2.3 to 3.0. #312

Add access to the event 
- cursorActivity 
- viewportChange 
- gutterClick 
- focus 
- blur 
- scroll 
- update

With the same way of  `onChange` : 

``` html
 <textarea 
ng-model="input"
ui-codemirror="{
   onChange:         doChange ,
   onCursorActivity: doSomething
}" 
>
```

Demo : http://jsbin.com/acitel/1/

### DIFF
--- a/modules/directives/codemirror/codemirror.js
+++ b/modules/directives/codemirror/codemirror.js
@@ -5,11 +5,12 @@
 angular.module('ui.directives').directive('uiCodemirror', ['ui.config', '$timeout', function (uiConfig, $timeout) {
 	'use strict';
 
+	var events = ["cursorActivity", "viewportChange", "gutterClick", "focus", "blur", "scroll", "update"];
 	return {
 		restrict:'A',
 		require:'ngModel',
 		link:function (scope, elm, attrs, ngModel) {
-			var options, opts, events, onChange, deferCodeMirror, codeMirror;
+			var options, opts, onChange, deferCodeMirror, codeMirror;
 
 			if (elm[0].type !== 'textarea') {
 				throw new Error('uiCodemirror3 can only be applied to a textarea element');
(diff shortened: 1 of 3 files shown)


modules/directives/codemirror/codemirror.js
```diff
@@ -2,77 +2,70 @@
...
-        if (codemirror) {
-          codemirror.toTextArea();
-        }
-        // Create the new codemirror widget
-        codemirror = CodeMirror.fromTextArea(elm[0], options);
-      };
 
-      // Initialize the code mirror widget
-      updateCodeMirror(uiCodemirrorGet());
+			events = "cursorActivity viewportChange gutterClick focus blur scroll update".split(' ');
```
ProLoser: Why not just make this an array to begin with? Also, might as well declare it up around line 7

douglasduteil: Yep. The same with `options` [line 18](https://github.com/angular-ui/angular-ui/pull/314/files#L0R18) ?

ProLoser: No, attrs isn't available earlier up in the execution chain, but the events array never changes and doesn't need to be created on every linking pass.

douglasduteil: Yes `opts` changes for each element 
line 19 :  

```
opts = angular.extend({}, options, scope.$eval(attrs.uiCodemirror));
```

But what about `options`  that depends of the `uiConfig`
 line 18:

```
options = uiConfig.codemirror || {};
```

Is it use whatever the element during compile-time ?

ProLoser:
Needed to review this more before I merged it.

Why are you deferring the initialization? You did not explicitly state in comments the reason for this, and it's not good to simply use this as an end-all solution needlessly.

douglasduteil:
Yep sorry I'll add some comments ...
Basically, I do it to let the compiler go through the whole node. When added CodeMirror takes textarea.nextSibling that isn't compile yet. 
From [codemirror.js#L3179](https://github.com/marijnh/CodeMirror/blob/master/lib/codemirror.js#L3179) : 

``` javascript
textarea.parentNode.insertBefore(node, textarea.nextSibling);
```

That why it's doing strange things (see http://jsbin.com/aqecex/6)

PR #328, opened 2013-01-01
Added an equal filter

bignall:
Added an equal filter, see description here: https://groups.google.com/forum/?fromgroups=#!category-topic/angular-ui/ui/7PLpGnHSZi8

### DIFF
--- a/modules/filters/equal/equal.js
+++ b/modules/filters/equal/equal.js
@@ -1,0 +1,27 @@
+/**
+ * Filters out objects that are not equal for the specified (key, value) pairs.
+ * @param [obj] { key: value, key: value, ... }
+ * if the param is empty, no filtering will be performed
+ * @return {array}
+ */
+
+angular.module('ui.filters').filter('equal', function () {
+
+    return function(items, obj) {
+        if (!angular.isArray(items) || !angular.isObject(obj)) return items;
+        if (Object.keys(obj).length == 0) return items;
+
+        var filtered = [];
+        angular.forEach(items, function(item) {
+            if (!angular.isObject(item)) return;
+            for (var key in obj) {
+                if (!angular.equals(item[key],obj[key])) return;
+            }
+            filtered.push(item);
+            return;
+        });
+        return filtered;
+
+    }
+});
+
--- a/modules/filters/equal/test/equalSpec.js
+++ b/modules/filters/equal/test/equalSpec.js
@@ -1,0 +1,55 @@
+describe('equal', function () {
+  var equalFilter;
+
+  beforeEach(module('ui.filters'));
+  beforeEach(inject(function ($filter) {
+    equalFilter = $filter('equal');
+  }));
+
+  it('should return the items if the items is not an array', function() {
+    var arrayToFilter = 'not an array';
+    expect(equalFilter(arrayToFilter)).toEqual(arrayToFilter);
+  });
+
+  it('should return the items if the obj is not an object', function() {
+    var arrayToFilter = ['item1', 'item2'];
+    var obj = 1;
+    expect(equalFilter(arrayToFilter, obj)).toEqual(arrayToFilter);
+  });
+
+  it('should return the items if the obj is empty', function() {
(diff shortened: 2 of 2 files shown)


ProLoser:
So a few thoughts:

Obviously this seems nearly identical to the existing filter-filter (confusing) except it only accepts exact matches, like some sort of strict option.

I am wondering, did you examine the source code to confirm that there isn't an additional undocumented argument?

Also, perhaps this should be a patch to the core? So that I could do `$filter('filter')(array, match, true); // true = strict`

Finally, I'm also curious, is it possible to _extend_ a filter? Like, simply name the new one 'filter' and theoretically, it SHOULD override the original one. However, if you inject the old filter into the new one, you can simply add a check for a strict parameter and if there is none then simply pass it through to the original filter?

This is theoretical however and I'm not sure it will work but I'd like to try first.

bignall:
I did examine the source code looking for an undocumented argument and didn't find anything. I thought about a patch to core and that's a good idea, so I'll pursue it.

Afaik it's not possible to extend a filter.  I didn't actually pursue that line though and it's a good idea so I'll try to check it out as well.

ProLoser:
I want to _test_ extending before we merge this.

Do me a favor and do `var filter = $filter('filter');` in the **outer** function and see if the reference is saved properly.

bignall:
Yes it does save the reference properly. However if you try to call the extending filter 'filter' instead of 'equal' then there is a circular dependency.  So you could use the filter 'filter' within your filter - not exactly extending it in the precise sense.

(And I definitely agree that filter-filter is very confusing ;)

ProLoser:
You know what I also want to add? Object filtering. Can only filter arrays.

bignall:
Okay, so here are two plunkers to look at.

http://plnkr.co/edit/rDPakO?p=preview

Shows that getting the reference to the filter can be saved used in the filter.  It adds a strict and !strict option to the equal filter.  When strict = false then it just passes the items and expression through to the $filter('filter').

http://plnkr.co/edit/vxZH8T?p=preview

Shows that you cannot rename the filter to filter as it causes a circular dependency:

```
Error: Circular dependency: filterFilter
    at Error (<anonymous>)
    at Object.c [as get] (http://ajax.googleapis.com/ajax/libs/angularjs/1.0.3/angular.min.js:26:9)
...
```

You'd have to keep a reference to the original filter somewhere outside of the filter you are creating.  I'm not sure where this would be appropriate.

Filtering objects.. hmm.. an interesting thought.  But what does it mean to filter an object?  Treat an object as an array and return a new object containing only the elements with values matching the expression?

pkozlowski-opensource:
@bignall can't you simply pass a function to the `filter` filter? 

I'm not sure we need a dedicated filter for the strict equality... It is so easy to pass a function if you need a custom comparison logic.... I don't know, I'm not thrilled by the idea of this filter... But those are only my 3 cents of course...

bignall:
Yes you can pass a function to the filter, and I did that for a while.  But I was using it so much and repeating the exact same code in so many different apps that it became much easier and so useful to have one equal filter and all I had to do was add that filter and call the equal-filter instead of filter-filter(with the function), angular-ui seemed the perfect place for it.  

As you suggested previously, I'm working on a patch to core Angular to add a strict option to the filter-filter.  I almost have it working, so maybe it's better to hold off on this patch or we'd be removing it later if my strict option gets accepted.

ProLoser:
Yeah I might have to agree with @pkozlowski-opensource but I definitely see the validity of your point. I think it would be better to propose this as a patch to the core.

Just link me to the bug ticket and I will be happy to +1 it! And since @pkozlowski-opensource is now on the core team we can hope it won't be _completely_ ignored lol.

pkozlowski-opensource:
Here is the ticket: https://github.com/angular/angular.js/pull/1760

But I'm not sure that having me on the core team helps here as I'm not big fan of this change :-)
But ok, let's see!

bignall:
I see you found the ticket before I even pointed you to it ;)  Thanks!

PR #341, opened 2013-01-11
getRemote directive

edinella:
# getRemote directive

Loads remote data in current scope.
Useful for example to load options to feed selects, from APIs.

See it in action here:
http://plnkr.co/edit/qhfaenJ4uR0ttJkdXkn2

### DIFF
--- a/modules/directives/getremote/README.md
+++ b/modules/directives/getremote/README.md
@@ -1,0 +1,28 @@
+# get-remote directive
+
+Loads remote data in current scope.
+Useful to allow selects to load options from APIs, for example.
+
+# Requirements
+
+- $http
+
+# Usage
+
+Include module and apply the directive to an element, with your desired rule
+
+## Rule syntax
+
+    path as alias
+    
+Where:
+ - path: is the path to your data JSON file or API
+ - alias: is the name of the variable in current scope you choosed to feed with data
+
+## Example
+
+    <select get-remote="/api/users as myUsers" ng-options="usr._id as usr.name group by usr.depto for usr in myUsers"></select>
+    
+or simply
+
+    <pre get-remote="/path/to/data.json as myData">{{myData}}</pre>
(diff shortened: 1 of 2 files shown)


pkozlowski-opensource:
@edinella out of curiosity: why would you like to load data from a directive as opposed to fetching nad exposing data on a scope from a controller? Could you share a code snippet showing how it makes things nicer / better / shorter?

edinella:
Surely, I will prepare a detailed example.
Meanwhile, have you seen it working on Plunker?
http://plnkr.co/edit/qhfaenJ4uR0ttJkdXkn2

pkozlowski-opensource:
yeh, so the plunk, but I'm getting bad gut feeling about it... For me it looks like you are moving $http logic to a template... What should happen in case of error? As it is now it will fail silently... And what about unit-testing...

Will let others comment but I'm somehow uneasy about this one....

petebacondarwin:
@pawel - I guess you could add attributes for success and error that
provided callback?

It is an interesting concept, but I generally do agree with Pawel, though,
I am not sure this saves much typing and in a big application one is likely
wanting to do stuff with the data when it arrives before it goes on the
scope.

Pete

pkozlowski-opensource:
@petebacondarwin Sure, this directive could be extended to provide callbacks and even data transformers. But IMHO it just moves code from JavaScript (where it is easy to test) to a template (where it is impossible? to test).

edinella:
Use case: http://plnkr.co/edit/32MRXD
To test the effect of a directive on DOM and Model don't seems too dificult, please see this: https://github.com/vojtajina/ng-directive-testing/blob/start/test/tabsSpec.js

ProLoser:
Sorry @edinella, I think the consensus is that this directive isn't appropriate for AngularUI.

However, don't let that dissuade you from submitting any other ideas you have!

PR #347, opened 2013-01-15
Enhancements to the passthrough directive

ProLoser:
Dropped uiChange, after doing some tests I confirmed that it cannot break anything by firing redundantly ('input' and 'change' events are always fired independently).

Added a uiRefresh property.
Added a uiDefer property/option for late-binding

### DIFF
--- a/modules/directives/jq/test/jqSpec.js
+++ b/modules/directives/jq/test/jqSpec.js
@@ -3,6 +3,7 @@ describe('uiJq', function () {
   scope = null;
   beforeEach(module('ui.directives'));
   beforeEach(function () {
+    jQuery.fn.foo = function () {};
     module(function ($provide) {
       $provide.value('ui.config', {
         jq: {foo: {}}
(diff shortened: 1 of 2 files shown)


ProLoser:
Need help getting unit tests to pass if anyone wants to take a look.

modules/directives/jq/test/jqSpec.js
```diff
@@ -1,47 +1,91 @@
...
+  });
+  describe('passing uiChange to options', function() {
+    it('should watch a form control change event and trigger an input event if true', function() {
+      var bar = false;
+      var element = compile('<input ui-jq="foo">')(scope);
+      element.bind('input', function(){
+        bar = true;
+      });
+      element.trigger('change');
+      expect(bar).toBe(true);
```
ProLoser, line 70: This test is failing and I can't figure out why. Isn't `trigger()` synchronous?

ajoslin, line 70: In https://github.com/angular-ui/angular-ui/blob/master/modules/directives/jq/jq.js#L35, you check if it has attrs.ngModel before binding to change events. If you add `ng-model="something"` to line 65 it works.

ProLoser, line 70: Damn my efficient logic.

PR #351, opened 2013-01-15
uiMapLocale pull-request

thiagofelix:
New directive **uiMapLocale** to make the map ( **uiMap** ) center at the same place where the user is, base on his geo localization. test unit included and a lib to mocking navigator.geolocation also included.

Reference: [geolocation-API](http://www.w3.org/TR/geolocation-API/)


modules/directives/maplocalization/maplocale.js
```diff
@@ -0,0 +1,53 @@
...
+ * @param ui-map-locale {object literal}  Hash of property:value to use as options during the geolocalization (optional)
+ *
+ * More about these options
+ * @{@link http://www.w3.org/TR/geolocation-API/#position-options}
+ */
+angular.module('ui.directives').directive('uiMapLocale', ['$parse', function($parse) {
+  var map, options;
+
+  var successCallback = function(position) {
+    console.log('Usuário foi geolocalizado com successo: ', position);
```
ProLoser: Can't have console logs in here.

ProLoser:
Is it worth it to have a separate directive just to center your location?

modules/directives/maplocalization/maplocale.js
thiagofelix: I forgot these console.log in the code, sorry, my bad.

ProLoser:
You need to remove your changes to the build files. You should only be committing source file changes.

ProLoser:
Pinging @joseym @ajoslin @petebacondarwin and anyone else interested to vote on this one. I'm not sure I see the need for creating a dedicated directive just to do this? I feel like it is more relevant as an option on the existing directive, and I worry about imposing a specific geolocation API.

joseym:
:-1: While I like the idea of being able to center the map to the users location this seems like it should be an option added to the existing **uiMap** directive.

ajoslin:
Yeah I don't think this is directive worthy.  Maybe option worthy, but I don't know about that either.  

## 

Andy Joslin
{ cell: 919-723-8919, email: andytjoslin@gmail.com }
Sent with Sparrow (http://www.sparrowmailapp.com/?sig)

Issue #352, opened 2013-01-15
Problem with ng-class and Select2


murilobr:
There is a problem with Select2 and ng-class, because ng-class is changed just inside select and not on the div created to emulate a select box.

Add to directive uiSelect2 the following code:

scope.$watch(attrs.ngClass, function(newVal, oldVal) {
          elm.prev().addClass(newVal);
          elm.prev().removeClass(oldVal);
        });

I will create a pull request.

ProLoser:
Unless you can provide an explicit use-case in which this would be necessary instead of simply putting the class on the container, I'm going to close this issue

PR #353, opened 2013-01-15
solution to issue #352

murilobr:
Solution to problem using ng-class with Select2.
When class changes just change select element, but not div.select2-container


ProLoser:
You need to add tests. Are you sure this works?

`ngClass` is not a simple class name, it's an evaluated statement that can contain a string or even an object.

murilobr:
@ProLoser you are correct I´ll verify it.

ProLoser:
@murilobr I'm going to close this PR for now. Feel free to reopen if/when you get it working.

PR #356, opened 2013-01-16
Add 'dist' task to gruntfile to specify build directory

ajoslin:
Add 'dist' task to gruntfile to specify build directory

### DIFF
--- a/grunt.js
+++ b/grunt.js
@@ -86,8 +86,6 @@ module.exports = function (grunt) {
         lessBuildFiles = lessBuildFiles.concat(moduleless);
       });
 
-      //Set config with our new file lists
-      grunt.config('builddir', 'build/custom');
       grunt.config('concat.build.src', jsBuildFiles);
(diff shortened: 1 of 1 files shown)


ajoslin:
Does this look good to everyone else?

petebacondarwin:
What happens if you don't specify dist ?

Pete
...from my mobile.

ajoslin:
If you don't specify, it uses the default specified in `initConfig`

petebacondarwin:
I can't see it specified in the initConfig. Perhaps we should add it

ajoslin:
It's easy to miss, it's the first line of initConfig https://github.com/angular-ui/angular-ui/blob/67eed39323f138cab22ef79a6f87df172be57c49/grunt.js#L11

ProLoser:
Looks good to me. Just out of curiosity, what were you doing that lead you to need this (not that I have a problem with it)?

ajoslin:
It's for the build server, to put files in a different folder.

PR #357, opened 2013-01-16
Update modules/directives/mask/mask.js

getuliojr:
This new version accepts a parameter to remove or not the mask as it is updated to the model and also make it more compatible with uiDate Controller

### DIFF
--- a/modules/directives/mask/mask.js
+++ b/modules/directives/mask/mask.js
@@ -1,18 +1,32 @@
 /*
  Attaches jquery-ui input mask onto input element
+   @param [ui-mask] {object} Options available: 
+   - mask: set the mask of the object
+   - removeMask: true or false, if false the model will retain the mask as it was typed by the user.
  */
-angular.module('ui.directives').directive('uiMask', [
-  function () {
+angular.module('ui.directives').directive('uiMask', ['ui.config',
+  function (uiConfig) {
     return {
       require:'ngModel',
       link:function ($scope, element, attrs, controller) {
-
+        //get options
+        var options = angular.extend({}, uiConfig.mask, $scope.$eval(attrs.uiMask));
+        
         /* We override the render method to run the jQuery mask plugin
          */
         controller.$render = function () {
           var value = controller.$viewValue || '';
           element.val(value);
-          element.mask($scope.$eval(attrs.uiMask));
+          //If a mask parameter was set as object, parse its value
+          if (options["mask"] != undefined) {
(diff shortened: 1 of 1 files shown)


pkozlowski-opensource:
@getuliojr thnx for your PR. Would you mind adding tests to expose what you are trying to fix with your change?

modules/directives/mask/mask.js
```diff
@@ -25,13 +39,20 @@ angular.module('ui.directives').directive('uiMask', [
...
-         */
-        element.bind('keyup', function () {
-          $scope.$apply(function () {
-            controller.$setViewValue(element.mask());
-          });
-        });
+        /* When changed, update the view value but only if it is not been used with the directive uiDate
+           uiDate will parse as a date so it is better format for dates, instead of string.
+        */
+        if ((attrs["uiDate"] == undefined) && (attrs["uiDateFormat"] == undefined)) {
```
ProLoser, line 45: There's no reason / need to check for uiDateFormat since this affects only the model value and won't / shouldn't be used without uiDate, right?

getuliojr:
UiDateFormat can be set out, the objective to make it more compatible with uiDate that will parse as DATE what is more expected to be used instead of STRING that this model would be using. I am setting a plnkr now to show it.

getuliojr:
Here is a working example of the changes, hope it can explain by itself

http://plnkr.co/edit/Dq14NVBCWdL5ui1cRzh9?p=preview

ProLoser:
1. You should probably be hooking into the `input` event instead of `change`
2. Why do you even need to hook into this event? Doesn't `$render` do this automatically?

getuliojr:
It was wired on keyup on original code to parse the values back to the model. I did not see a reason to be listening to keyup so instead I changed to bind to 'change' event to parse the values back to the model.

ProLoser:
Does the mask plugin preventDefault() or something? Why is it necessary to explicitly set the model?

## 

Dean Sofer
DeanSofer.com
714.900.2254

getuliojr:
First I did not want to make a huge rewrite in the diretive in order to not brake it, since it is the first time I am contributting and as so I need to take small steps at a time.

Second, since it you might set the model value with or without a mask you can´t have it get the defaultvalue set. You do need to check this step.

But I do think preventDefault might be been used, since it is a plugin. Did not check it though.

shaungrady:
uiMask has since been refactored, this issue is no long applicable. It will be in the next AngularUI release.

PR #358, opened 2013-01-17
uiModal onShown and onHidden function callbacks, updated testing of modal closing

cymantic:
...ring of shown and hidden events by bootstrap-modal.js

wire up by adding attribute value to ui-modal
e.g. ui-modal='{onShown:shownFn,onHidden:hiddenFn}'

updated ui-modal spec to make sure that the close tests take place after the modal has been opened first, otherwise they only confirm that the ui-modal hasn't been opened.


ProLoser:
I'm not sure I see the purpose of this when you can simply do ui-event instead.

cymantic:
No ui-event at that point... (or didn't know about it) :-)

PR #369, opened 2013-01-21
Handle invalid masked setups more gracefully.

sja:
If there was a setup problem, i.e. you have <input ui-mask="99">, maskedinput stops here, because input.data($.mask.dataName) is undefined.

### DIFF
--- a/test/lib/maskedinput/jquery.maskedinput-1.3.js
+++ b/test/lib/maskedinput/jquery.maskedinput-1.3.js
@@ -54,7 +54,12 @@
 		mask: function(mask, settings) {
 			if (!mask && this.length > 0) {
 				var input = $(this[0]);
-				return input.data($.mask.dataName)();
+				var maskFn = input.data($.mask.dataName);
+				if (typeof maskFn !== 'function') {
+					console.error('maskedinput was not initialized correctly');
+					return;
+				}
+				return maskFn();
 			}
 			settings = $.extend({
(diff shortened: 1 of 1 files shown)


petebacondarwin:
Hi @sja - Thanks for the PR.  In what situation does this occur?  I would have thought that if the mask does not initialize properly then that is a programming error that should not be hidden in a console message.

sja:
Well, you're right, thats a programming error. I see three solutions:
1. Just leave a comment with a hint and try to call this undefined.
2. Do not call undefined and just return. This results is that only nothing happens. A programmer have to step through this (from his perspective) external code.
3. Do a console.error to force the programmer to see this point in code and immediately know what to do.

What do you prefer? Every solution has it's pros and cons.

Sebastian

Am 21.01.2013 um 14:52 schrieb Pete Bacon Darwin notifications@github.com:

petebacondarwin:
Or throw an exception with a useful message?

```
  var maskFn = input.data($.mask.dataName);
  if (typeof maskFn !== 'function') {
    throw new Error('The Mask widget is not correctly set up');
  }
  return maskFn();
```

sja:
You're right, that would be the best. Please see the new PR: https://github.com/angular-ui/angular-ui/pull/370

pkozlowski-opensource:
+1 for throwing an error. Log in the console could go unnoticed.

PR #370, opened 2013-01-21
Handle invalid masked setups more gracefully.

sja:
If the mask widget was not set up correctly, throw an Error with a hint.

### DIFF
--- a/modules/directives/mask/test/maskSpec.js
+++ b/modules/directives/mask/test/maskSpec.js
@@ -35,4 +35,13 @@ xdescribe('uiMask', function () {
   describe('model binding on ui change', function () {
     //TODO: was having har time writing those tests, will open a separate issue for those
   });
+
+  describe('should fail', function() {
+    it('errors on missing quotes', function() {
+      $rootScope.x = 42;
+      var errorInputHtml = "<input ui-mask=\"(9)9\" ng-model='x'>";
+      element = $compile(errorInputHtml)($rootScope);
+      expect($rootScope.$digest()).toThrow();      
+    });
+  });
 });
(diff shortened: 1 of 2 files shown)


petebacondarwin:
Great!  Now we just need a unit test for it!

sja:
Ok, I'm working on it. My problem is, that I start `grunt server` and it stops with listing `...templateSpec.js`. After that, it should report a running server, but nothing happens.
It worked a moment ago. I reverted my changes but it doesn't come up again.
Meta Info: OS X 10.8, fresh installed npm-deps, node 0.8.16.

ProLoser:
So it took me a little while to figure out how to do tests with our build. It's a 2 step process:
1. Run `grunt server`
2. Open a browser and point it to localhost:8080 (or whatever the grunt-server says)
3. Open another terminal and run `grunt test` (twice to be safe)

If your tests hang, bring focus to the web page. It's a bug with the ui-calendar.

ProLoser:
@petebacondarwin can you give this another once-over so I / you can go ahead and merge this one?

modules/directives/mask/test/maskSpec.js
```diff
@@ -35,4 +35,13 @@ xdescribe('uiMask', function () {
   describe('model binding on ui change', function () {
     //TODO: was having har time writing those tests, will open a separate issue for those
   });
+
+  describe('should fail', function() {
+    it('errors on missing quotes', function() {
+      $rootScope.x = 42;
+      var errorInputHtml = "<input ui-mask=\"(9)9\" ng-model='x'>";
+      element = $compile(errorInputHtml)($rootScope);
+      expect($rootScope.$digest()).toThrow();
```
petebacondarwin, line 44: I think you should be passing a function to `expect()` here:

```
expect($rootScope.$digest).toThrow();
```

And if we wanted to be really particular this could be:

```
expect($rootScope.$digest).toThrow('The Mask widget is not correctly set up');
```

PR #372, opened 2013-01-22
Updated Calendar to Utilize uiConfig for Event Callbacks 

joshkurz:
The calendar was utilizing an isolated scope to take in the events. This was not good for communication between the calendar and the scope's event objects. We need to evaluate the events from the ng-model attribute so that scope functions can be wired to the calendar's event callbacks.

There has been one major change to the api of the calendar. We are now taking in the eventSources through ngModel so that an event array of objects and arrays of source objects can be added at the same time. 

Now the events object that is added into ng-model needs to be structured as a nested array. This is all documented here http://arshaw.com/fullcalendar/docs/event_data/eventSources/

We are still requiring ng-model so that the calendar cannot be created without an events array of individual event objects. 

We are detecting whether or not the user has created header options for the calendar and if so we use those instead of the defaults.

### DIFF
--- a/modules/directives/calendar/calendar.js
+++ b/modules/directives/calendar/calendar.js
@@ -2,55 +2,61 @@
 *  AngularJs Fullcalendar Wrapper for the JQuery FullCalendar
 *  API @ http://arshaw.com/fullcalendar/ 
 *  
-*  Angular Calendar Directive that takes in the eventSources nested array object as the ng-model and watches an object named events for changes, 
+*  Angular Calendar Directive that takes in the eventSources nested array object as the ng-model and watches (eventSources.length + eventSources[i].length) for changes, 
 *  to update the view accordingly. Can also take in multiple event urls as a source object(s) and feed the events per view.
 *
 *
-*  The calendar must have an object named events on the scope to update itself properly when the events array is altered from outside of the calendar.  
+*  The calendar will watch any eventSource array and update itself when a delta is created  
 *
 */
 
 angular.module('ui.directives').directive('uiCalendar',['ui.config', '$parse', function (uiConfig,$parse) {
-    uiConfig.uiCalendar = uiConfig.uiCalendar || {};       
+     uiConfig.uiCalendar = uiConfig.uiCalendar || {};       
      //returns calendar     
      return {
         require: 'ngModel',
         restrict: 'A',
-         link: function(scope, elm, attrs) {
-            //update the calendar with the correct options
-            function update() {
-            var view;
-            if(elm.fullCalendar('getView')){
-              //setting the default view to be whatever the current view is. This can be overwritten. 
-              view = elm.fullCalendar('getView').name;
-            }
-            //If the calendar has options added then render them.
-            var expression,
-              options = {
-                defaultView : view,
-                eventSources: scope.$eval(attrs.ngModel)
-              };
-            if (attrs.uiCalendar) {
-              expression = scope.$eval(attrs.uiCalendar);
-                //if header has not been included then render default header options
-                if(!expression.hasOwnProperty('header')){
-                  expression.header = {
-                    left: 'prev,next today',
-                    center: 'title',
-                    right: 'month,agendaWeek,agendaDay'
-                  };
+          link: function(scope, elm, attrs, $timeout) {
+            var sources = scope.$eval(attrs.ngModel);
+            var tracker = 0;
+            /* returns the length of all source arrays plus the length of eventSource itself */
+            var getSources = function () {
+              tracker = 0;
+              angular.forEach(sources,function(value,key){
+                if(angular.isArray(value)){
+                  tracker += value.length;
                 }
-             } else {
-              expression = {};
-             }
-             angular.extend(options, uiConfig.uiCalendar, expression);
-             elm.html('').fullCalendar(options);
+              });
+              return tracker + sources.length;
+            };
+            /* update the calendar with the correct options */
+            function update() {
+              scope.calendar = elm.html('');
+              var view = scope.calendar.fullCalendar('getView');
+              //calendar object exposed on scope
+              if(view){
+                view = view.name; //setting the default view to be whatever the current view is. This can be overwritten. 
+              }
+              /* If the calendar has options added then render them */
+              var expression,
+                options = {
+                  defaultView : view,
+                  eventSources: sources
+                };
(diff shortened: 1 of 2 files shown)


modules/directives/calendar/calendar.js
```diff
@@ -1,61 +1,54 @@
...
+                eventSources: scope.$eval(attrs.ngModel)
+              };
+            if (attrs.uiCalendar) {
+              expression = scope.$eval(attrs.uiCalendar);
+                //if header has not been included then render default header options
+                if(!expression.hasOwnProperty('header')){
+                  expression.header = {
+                    left: 'prev,next today',
+                    center: 'title',
+                    right: 'month,agendaWeek,agendaDay'
```
ProLoser: Why are you hard-coding the header and why are you only doing it here?

modules/directives/calendar/calendar.js
```diff
@@ -1,61 +1,54 @@
...
-              }
-              else{
-                expression = {};
-              } 
-              //extend the options to suite the custom directive.
-              angular.extend(options, uiConfig.uiCalendar, expression);
-              //call fullCalendar from an empty html tag, to keep angular happy.
-              elm.html('').fullCalendar(options);
+            var view;
+            if(elm.fullCalendar('getView')){
```
ProLoser: Can probably memo-ize this method call (trivial)

modules/directives/calendar/calendar.js
joshkurz: Im not. If the user does not specify any header options then we will go with these defaults, because they are common, but the user can overwrite the header in uiConfig. All im doing is Just allowing the user to not have to put this into uiConfig if they dont want to. This is the most common setup for the calendar. If they just put header: {center: 'title'} that is all that will be entered into the options.

modules/directives/calendar/calendar.js
joshkurz: <h4 style='font-family:;'>Like So?</h4>
    <pre class="bold" style='font-family:;'>
    if(elm.fullCalendar('getView')){ view = elm.fullCalendar('getView').name; }</pre>

ProLoser: By doing:

``` js
view = elm.fullCalendar('getView')
if (view)
  view = view.name;
else
  view = false;
```

Or something like that. You just don't have to call the method twice (not that it's a huge overhead or something) I'm being OCD. You can ignore me.

joshkurz: Oh yeah. We don't even need the else statement. If view is left as undefined then the calendar renders 'month' as the default.

modules/directives/calendar/calendar.js
ProLoser: Again my question is doesn't fullCalendar render this by default if you don't pass a header option?

ProLoser: http://arshaw.com/fullcalendar/docs/display/header/ according to the docs that object is both inaccurate and also used by default.

Why don't you just delete this header-specific code? It doesn't really seem to serve any real additional purpose over fullCalendar's native behavior.

I prefer to not override plugin defaults.

joshkurz: ok. You got me on that one.

ProLoser:
What's the status of this PR?

joshkurz:
on it. will have an update later this evening.

On Wed, Jan 30, 2013 at 6:15 AM, Dean Sofer notifications@github.comwrote:

> What's the status of this PR?
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/angular-ui/angular-ui/pull/372#issuecomment-12884963.

## 

_Josh Kurz_

ProLoser:
If you can there's a new AngularUI bug on Stack Overflow about the calendar plugin too.

## 

Dean Sofer
DeanSofer.com
714.900.2254

joshkurz:
Not sure what happened on the test situation here I don't think they ever ran.

ProLoser:
You really shouldn't do PRs on your master fork. I'm actually curious if that's why Travis ignored it?

ProLoser:
Getting a merge conflict with this:

``` js
++<<<<<<< HEAD
 +            //watching the length of the array to create a more efficient update process. 
 +            scope.$watch( 'events.length', function( newVal, oldVal )
 +            {
 +              //update the calendar on every change to events.length
 +              update();
 +            });
 +        }
++=======
+               /* watches all eventSources */
+               scope.$watch(getSources, function( newVal, oldVal )
+               {
+                 update();
+               });
+          }
++>>>>>>> 2531e964d4b032c2d1aae126250c0e2105c88abf
```

joshkurz:
ahhhhhhhhhhhhh. Wow. thats lame.

On Sat, Feb 2, 2013 at 4:38 PM, Dean Sofer notifications@github.com wrote:

> Getting a merge conflict with this:
> 
> ++<<<<<<< HEAD
> -            //watching the length of the array to create a more efficient update process.
> -            scope.$watch( 'events.length', function( newVal, oldVal )
> -            {
> -              //update the calendar on every change to events.length
> -              update();
> -            });
> - ```
>      }++=======+               /* watches all eventSources */+               scope.$watch(getSources, function( newVal, oldVal )+               {+                 update();+               });+          }++>>>>>>> 2531e964d4b032c2d1aae126250c0e2105c88abf
>   ```
>   
>   —
>   Reply to this email directly or view it on GitHubhttps://github.com/angular-ui/angular-ui/pull/372#issuecomment-13037880.

## 

_Josh Kurz_

ProLoser:
Is there supposed to be an `update()` right before the watch as well as inside it? Also I'm assuming I should use the `getSources` version?

ProLoser:
Can you please merge the latest version of ui/master onto your fork so that I don't have to resolve the conflicts?

joshkurz:
that update helps some of the tests pass. I will try it without it and see
what happens now that the calendar is much more grown up. I will resolve
all the conflicts in a byte. Give me a tad bit of time. SuperBowl Sunday u
know.

On Sun, Feb 3, 2013 at 2:26 AM, Dean Sofer notifications@github.com wrote:

> Can you please merge the latest version of ui/master onto your fork so
> that I don't have to resolve the conflicts?
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/angular-ui/angular-ui/pull/372#issuecomment-13043542.

## 

_Josh Kurz_

Issue #263, opened 2012-11-15
Select2 doesn't show initial values


MykolaBilyi:
Select2 doesn't show initial values
This happens because it sets up the value before the actual plugin initialisation

http://plnkr.co/edit/Wckzzv?p=preview

bvkimball:
I have this issue too.  I did this to fix it.  Its not completely right but should get you going in the right direction

```
      if(opts.initSelection) {
          var initSelection = opts.initSelection;
          opts.initSelection = function(element, callback) {
            initSelection(element, function(value) {
              controller.$setViewValue(value);
              callback(value);
            });
          };
        }
       // ADDED THIS ELSE STATEMENT 
        else {
          opts.initSelection = function(element, callback) {
              callback(controller.$modelValue);
          };
        }
```

ProLoser:
This looks like it would be the proper fix. You should open a pull request
with the patch.

ni-ka:
+1 the code from  bvkimball looks good, using it and it works

philippd:
+1 i am also using bvkimballs fix and it's working

jph00:
After applying that pull request, I have a bug if my ng-model references an id instead of an object. In the scope.$watch we have elm.select2('data', controller.$viewValue), but if ng-model refers to an id this needs to be elm.select2('val', controller.$viewValue).

At the moment, it's still very hard to get a basic 'edit' form working when in ajax mode and with minimumInputLength set, since the select2 will not display the initial value without quite a bit of work (i.e. setting initSelection to an appropriate function, and making the change above).

ProLoser:
0414834a353f88e6dde41541c2586cad4a1237ea

Issue #228, opened 2012-10-14
ui-select2 with no initial options


daryllstrauss:
I had a controller that needed to initialize a ui-select2 from data that would be fetched from a service. The controller basically did this:

```
`$scope.allprojects=Project.query({'scope':'all'});`
```

The partial was written like this:

  &lt;select ui-select2 ng-model='project'&gt;
    &lt;option ng-repeat='proj in allprojects' value="{{proj._id}}"&gt;{{proj.name}}&lt;/option&gt;
  &lt;/select&gt;

It had strange inconsistent results. Sometimes it worked fine. 
1) The menu would be empty to start and would be wide enough to hold my widest project string. 
2) The menu would default to the first item in the project list.
3) The model would track changes in the ui-sliders.

Most of the time it would fail:

1) The menu would be very narrow, just showing me the triangle.
2) There was no item selected at startup.
3) If I opened the ui-select2 I would see the list of projects, and I was able to select them, but my model wouldn't update.

I worked around the problem by adding a:
<option value="">Project</option>
choice before the repeating clause.
That would initialize to the word "Project", but I noticed the menu was only as wide as the word "Project" not the widest choice. If I selected a wider choice the menu would resize. The model would update correctly as the menu changed.

I hope that's enough information for you to diagnose with.

ProLoser:
For width calculations, you have to sort it out yourself. Use explicit dimensions or mess with Select2's width calculation property.

As for initial value, there is no reason why any option should be selected by default (assuming the value of your ng-model is empty). If you want to work with initial values, make sure you set the value to your ng-model. Remember that you can set it to the first value of the response object once it completes its request.

daryllstrauss:
My point is that the behavior is very different depending on whether the projects has returned from it's ajax call. The model not updating is the biggest problem, but it looks like it short circuits some initialization if the initial select is empty.

ProLoser:
If you're not using it the way it's intended to be used, it's hard for me to account for what will happen.

http://en.wikipedia.org/wiki/Race_conditions

ProLoser:
Have you tried adding an empty `<option>`?

daryllstrauss:
On 10/13/2012 07:56 PM, Dean Sofer wrote:

> Have you tried adding an empty |<option>|?
> 
> —
> Reply to this email directly or view it on GitHub 
> https://github.com/angular-ui/angular-ui/issues/228#issuecomment-9415411. 
> 
[...]

PR #373, opened 2013-01-22
fix(select2) Can now set ngModel value to simple objects

### LINKED ISSUE #135: select2 directive value setting doesn't work
requirements for usage of select2
- it should work  in a kind of tagging mode. this means you should be able to add on the fly.
- working with an object array (e.g. id,name object) 

here is a fiddle: http://jsfiddle.net/maku/7AvKZ/
it shows that you are able to set the value after a certain amount of time (via click action).
but when the value is already set at the beginning in the controller it doesn't work correctly.

See also https://groups.google.com/forum/#!topic/angular-ui/Yws0P0-Un2c

ProLoser:
An expanded fix from #309 

Possibly Fixes #135 #263 #228

### DIFF
--- a/modules/directives/select2/select2.js
+++ b/modules/directives/select2/select2.js
@@ -5,7 +5,7 @@
  *     This change is so that you do not have to do an additional query yourself on top of Select2's own query
  * @params [options] {object} The configuration options passed to $.fn.select2(). Refer to the documentation
  */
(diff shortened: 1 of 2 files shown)


ProLoser:
Needs tests, this demos most test-cases http://plnkr.co/edit/Ni4QubpSRpfYcLgm4QXd (except multiples)

Issue #348, opened 2013-01-15
Route match directive


ProLoser:
So one directive that I need to write down because it keeps rolling around in my head is a route-matching directive for navigation highlighting.

It would come in handy when people use tabs or need to simply highlight the current page in a navigation bar.

Perhaps something like 

``` html
<li ui-route="#/somepage/*">
<li ui-route="index.html">
<li ui-route="details.html#/tab1">
<li ui-route="/tab{{ $index + 1 }}">
```

And other such nonsense. It would simply add some sort of 'active' class and we could continue to expand the syntax support yet still try to keep it simple (regex may be too powerful?)

Should also probably have a better directive name than 'route'

petebacondarwin:
Nice simple idea.

pkozlowski-opensource:
Yeh, +1 We could use it for bootstrap as well.

ProLoser:
Does this just toggle a class or do we do more with it?

capaj:
Right now I am just using ng-switch on $routeParams, so yeah, this would be definately helpful.

ProLoser:
Okay so you want more than just a simple class being toggled?

@capaj What does your switch look like? I don't know how this is directly related

Maybe it should be a rootscope method for the most flexibility?

`ng-class=" { active : match( '/someRoute' ) } "`

sudhakar:
Currently I am using something like,

``` javascript
    $scope.$watch('$location.path()', function (path) {
      $scope.activeNavId = path || '/';
    });

    /**
     * getClass compares the current url with the id.
     * If the current url starts with the id it returns 'active'
     * otherwise it will return '' an empty string. E.g.
     *
     *  // current url is '/products/1'
     *  getClass('/products'); // returns 'active'
     *  getClass('/orders'); // returns ''
     */
    $scope.getClass = function (id) {
      return ($scope.activeNavId.substring(0, id.length) === id) ? 'active': '';
    };
```

with bs markup

``` html
<ul class="nav nav-tabs nav-stacked">
  <li ng-class="'getClass(/page/edit)'">
    <a ng-href="/pages/edit/page.id">Page Details</a>
  </li>
</ul>
```

Problems that I am facing with this approach
1. If i have a path like '/pages/:pageid/details' & '/pages/:pageid/campaigns', then something like ng-class="'getClass(/pages/{{page.id}}/details)'" doesnt seem to work.
2. Too noisy as I have to specify route both in li tag and in ng-href

[...]

PR #374, opened 2013-01-22
feat(route) prototyping a new route checking directive

ProLoser:
In response to #348 

This is just a WIP, **DO NOT MERGE YET**

### DIFF
--- a/modules/directives/route/route.js
+++ b/modules/directives/route/route.js
@@ -1,7 +1,7 @@
 /**
  * Set a $activeRoute boolean to see if the current route matches
  */
-angular.module('ui.directives').directive('uiRoute', ['ui.config', '$location', function (uiConfig, $location) {
+angular.module('ui.directives').directive('uiRoute', ['$location', function ($location) {
   return {
     restrict: 'AC',
     priority: 100, // must occur before attrs.ngHref is removed
@@ -19,7 +19,7 @@ angular.module('ui.directives').directive('uiRoute', ['ui.config', '$location', 
       if (attrs.uiRoute) {
         attrs.$observe('uiRoute', function(newVal) {
           watcher = function watchRegex() {
-            var regexp = new RegExp('^' + watcher + '$', ["i"]);
+            var regexp = new RegExp('^' + newVal + '$', ['i']);
             $scope.$routeActive = regexp.test($location.path());
           };
           watcher();
--- a/modules/directives/route/route.js
+++ b/modules/directives/route/route.js
@@ -1,26 +1,31 @@
 /**
- * Set a $activeRoute boolean to see if the current route matches
+ * Set a $uiRoute boolean to see if the current route matches
  */
 angular.module('ui.directives').directive('uiRoute', ['$location', function ($location) {
   return {
     restrict: 'AC',
     priority: 100, // must occur before attrs.ngHref is removed
+    scope: false,
     link: function ($scope, elm, attrs) {
       var watcher = angular.noop;
 
       // Used by href and ngHref
       function observeHref(newVal) {
+        if ((hash = newVal.indexOf('#')) > -1)
+          newVal = newVal.substr(hash + 1);
         watcher = function watchHref() {
-          $scope.$routeActive = ($location.path().indexOf(newVal) > -1);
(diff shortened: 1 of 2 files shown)


modules/directives/route/route.js
```diff
@@ -0,0 +1,42 @@
...
+
+      // Used by href and ngHref
+      function observeHref(newVal) {
+        watcher = function watchHref() {
+          $scope.$routeActive = ($location.path().indexOf(newVal) > -1);
+        };
+        watcher();
+      }
+
+      if (attrs.uiRoute) {
```
sudhakar: attrs.uiRoute evaluates to false all the time. I guess its because its getting treated as directive attribute.

ProLoser: Hmm... you're right, but it's only if you use `{{}}` and even if you put text OUTSIDE of the brackets it's still undefined initially. So... why?

ProLoser: The only way around it I can think of is `elm.attr(attrs.$attr.uiRoute)` instead but this won't work if used as a class. I have NO idea why it's undefined.

sudhakar: Quickly groked ng-href directive. Even there, they are only observing the change on ng-href attr. In our case attrs.ngHref is available because we have set priority to 100.

Is it something to do with uiRoute and its expression are treated separately but with same priority 100. then in tat case order of execution is unknown (this is a wild wild guess..)

modules/directives/route/route.js
```diff
@@ -0,0 +1,58 @@
+/**
+ * Set a $uiRoute boolean to see if the current route matches
+ */
+angular.module('ui.directives').directive('uiRoute', ['$location', function ($location) {
+  return {
+    restrict: 'AC',
```
ajoslin, line 6: probably should restrict A

ProLoser:
@sudhakar and @ajoslin thanks for your help, this is finally ready to go!

sudhakar:
@ProLoser Looks good to me & much cleaner than my initial proposal..

ajoslin:
Landed in 0f65bd921975ca4345ae94ba9e5b81ec3adea11b

Narretz:
What does it do exactly? Are there docs for it?

ProLoser:
Not really lol. I should expand the docblock. I was going to add a demo when I push the new build.

@narretz Want to help me out an open a PR?

### Here's basically how it works:
1. Put `ui-route` where you want to do something depending on what the current route is (such as setting an active class on a nav link or hiding something)
2. Declare a route-match pattern
3. [optional] Specify an ng-model to store the boolean into, otherwise it will store the boolean into $uiRoute
4. Use this boolean to perform your behavior: `<a ng-class="{active:$uiRoute}">` or `<div ng-hide="$uiRoute">`
   - `$uiRoute` (or your specified model) is added to the scope and can be accessed from children

### Declaring a route-match pattern

**ui-route**
Support regex and `{{}}`

``` html
<li ui-route="/dashboard">
or
<li ui-route="/user/[0-9]*">
or
<li ng-repeat="tab in tabs" ui-route="/{{tab}}">
or
<li ng-repeat="tab in tabs" ui-route="/{{tab}}(/[0-9]*)?">
```

**ng-href**
Supports `{{}}` statically

``` html
<a ng-href="/dashboard" ui-route>
or
<a ng-href="/user/{{userId}}" ui-route>
```

**href**
Static only

``` html
<a href="/dashboard" ui-route>
```

### Working with the flag

``` html
<li ui-route="/dashboard" ng-class="{active:$uiRoute}">
  <a href="/dashboard">Dash</a>
  <ul>
    <li ui-route="/dashboard/stats" ng-model="isStatsRoute">
     <div ng-show="isStatsRoute && $uiRoute">
```

Yes it's a shitty example

ProLoser:
@joseym Again, I'm not sure how much you're willing to take on, but if you are willing to help me out, The above comment ^^ needs to be turned into an open PR for the demo page that we can merge in once we push v0.4.0

It does not need to have an updated version of angular-ui in the PR as that will be changed already.

joseym:
Sure, this is great.

PR #379, opened 2013-01-24
Fixing bug with deferred enhancement

ProLoser:
So I was a moron 2-fold:
- I forgot to remove part of the code that checked if `uiDefer=true` despite removing the other half
- `$scope.$evalAsync()` fires **before** rendering, not after.

I switched over to $timeout but now my tests are back to being a little bitch. Need help with tests.


ajoslin:
Try using `$timeout.flush()` to synchronously end all timeouts in the test.

ProLoser:
@ajoslin that fixed mosts tests (thanks) but now my 'must run after rendering has finished' test is failing. If I `flush()` does that occur **before** rendering?

PR #381, opened 2013-01-24
Update modules/directives/date/date.js

getuliojr:
When a date object is passed in the model instead of string and ui-date-format is set it was not parsing the value. With this patch it parses strings or return the date format to be set in the ui-date

### DIFF
--- a/modules/directives/date/date.js
+++ b/modules/directives/date/date.js
@@ -83,6 +83,8 @@ angular.module('ui.directives')
         modelCtrl.$formatters.push(function(value) {
           if (angular.isString(value) ) {
             return new Date(value);
+          } else if (angular.isDate(value)) {
+             return value;
           }
         });
         modelCtrl.$parsers.push(function(value){
@@ -96,6 +98,8 @@ angular.module('ui.directives')
         modelCtrl.$formatters.push(function(value) {
           if (angular.isString(value) ) {
             return $.datepicker.parseDate(format, value);
+          } else if (angular.isDate(value)) {
+             return value;
           }
         });
         modelCtrl.$parsers.push(function(value){


petebacondarwin:
@getuliojr - thanks for this PR but the whole point of date-format is that it handles the situation where the model is a string. If your model is a Date object then you shouldn't be using date-format since date will accept the object directly.

ProLoser:
@petebacondarwin you said there was an explicit problem trying to detect if it was an object and falling back, right? Is there a major reason this behavior can't be supported?

petebacondarwin:
No programming problem but it muddies the water and there was enough confusion about this before.

It is now very clear: date only accepts a date object.  If you want to pass a string then you provide date-format with a string that specifies what format it is in.

I would suggest that it is a programming error if you can't tell in your own app whether your model is going to be a string or an object.

ProLoser:
Iunno, I still vote to merge this in.

## 

Dean Sofer
DeanSofer.com
714.900.2254

petebacondarwin:
@getuliojr - Can we see the use case for this change?

ProLoser:
Pinging @pkozlowski-opensource, @joseym and @ajoslin on this one.

getuliojr:
Well thinking it better through I only see 2 cases:
1. You will use ui-date with ui-dateformat to parse string, that worked already and if you would have a date object ui-date would work already either. So the patch for this case is really unnecessary.
2. You don´t want to use ui-date, but you want to use ui-mask and format the date to be parsed correctly to ui-mask. In that case ui-date-format could be useful, but the code submitted would need a patch for this, since this version is not formatting the date, missed that part.

Based on that the patch submitted might really not be usefull and is just more code to maintain so should not merge. And a case to be used with ui-mask could be studied if is necessary, and the cases and if it can help submitted a version that will work with that case.

ProLoser:
I have done this myself in the past assuming that both variations were still supported.

I just thought of a major reason to merge this:

If you setup dateFormat in ui.config, then you are forcing dateFormat to be enabled for every single ui-date implementation, and there is no way to override it to be disabled. This means that if you want to setup a global config for uiDateFormat, you are stuck with setting values as strings across the board.

Again, I think that's a big reason to merge this in so I am not needlessly restricted. If there was confusion, we should simply be sure to clarify in the docs.

ajoslin:
My first impression is why not merge this?  It allows the developer to do the intuitive thing and doesn't mess up any current code.

petebacondarwin:
@ProLoser - what you said is not correct.  Whatever you do with ui.config the uiDateFormat directive is only triggered if you add it to an input that contains a uiDate directive.

@ajoslin - There is a very good reason not to include this.  With this change we are back to the situation where if you put a Date object in your model, it will be quietly converted back to a string.  Which is not very intuitive.  It was a right pain getting to the point where the directives did two very clear things.  Look back at all the issues around this before when people were confused about when you would have an object and when you can have a string.

joseym:
If this PR adds another method for the user, without breaking the current implementation why not merge? If it adds a layer of confusion then yes, that's potentially a problem, but as @ProLoser pointed out - the docs could be updated to note this.

petebacondarwin:
Because adding a layer of confusion is not a recommended library design
practice?
There was a lot of confusion last year around date <-> string conversion
that led to this date-format directive in the first place.
What is the use case?
Pete

PR #395, opened 2013-02-01
fix(ieshiv): Removed browser version check and closes #392

### LINKED ISSUE #392: IE Shiv modifications
We should simplify the shiv and potentially relocate it to it's own module.

Also, I think that there is no point in trying to do IE browser detection INSIDE the shiv. People should just be instructed to use conditional comments.

ProLoser:
Since it's impossible to load this asset asynchronously there isn't really a reason to put the checks in js either.


petebacondarwin:
LGTM.  Do we have any tests for this shiv?

ProLoser:
No. This occurred to me too. I didn't know if I should do this now or if I should wait until we refactored our project structure so that I would have a sensible location to throw tests into.

PR #405, opened 2013-02-06
Re-Wrote Calendar to Watch All Event Sources

joshkurz:
Many updates to the calendar have ensued in the last month. 
1. The calendar is attached to the scope so that any method can be called from a controller at any time.
2. We are using the eventSources field instead of events, so that more than one source obj can be added to the calendar at once.
3. Every Array length inside of eventSources is being watched for changes. 
4. The view is updated to be the current view on every update, so that the calendar does not switch back to the default view any time the watch is fired.
5. equalsTracker attrs has been added to force the calendar to update on use cases that render the total eventSources length equal.
6. Rewrote and added tests to accompany each breaking change.

### DIFF
--- a/modules/directives/calendar/calendar.js
+++ b/modules/directives/calendar/calendar.js
@@ -1,63 +1,66 @@
 /*
 *  AngularJs Fullcalendar Wrapper for the JQuery FullCalendar
-*  inspired by http://arshaw.com/fullcalendar/ 
+*  API @ http://arshaw.com/fullcalendar/ 
 *  
-*  Basic Angular Calendar Directive that takes in live events as the ng-model and watches that event array for changes, to update the view accordingly. 
-*  Can also take in an event url as a source object(s) and feed the events per view. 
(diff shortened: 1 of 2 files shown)


ProLoser:
@joshkurz Can you create a PR on the demo page with the necessary documentation changes that I can easily merge in when I push v0.4.0?

joshkurz:
Sure Thing.

joshkurz:
sure no problem. Ill even throw in some nice css to boot.

On Sat, Feb 9, 2013 at 5:26 AM, Dean Sofer notifications@github.com wrote:

> @joshkurz https://github.com/joshkurz Can you create a PR on the demo
> page with the necessary documentation changes that I can easily merge in
> when I push v0.4.0?
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/angular-ui/angular-ui/pull/405#issuecomment-13329280..

## 

_Josh Kurz_

PR #414, opened 2013-02-10
Add the jQuery UI Autocomplete wrapper

mirontoli:
Hi! I have seen many implementations of jQuery UI Autocomplete in angular UI, but there are no complete implementations. Here is the code for the autocomplete which can use all the options you can pass into a jQuery UI autocomplete. See my blog post for details: [angular jQuery UI autocomplete](http://sharepointkunskap.wordpress.com/2013/01/04/angular-jquery-ui-autocomplete/)

### DIFF
--- a/modules/directives/autocomplete/autocomplete.js
+++ b/modules/directives/autocomplete/autocomplete.js
@@ -1,0 +1,31 @@
+/*global angular */
+/*
+ jQuery UI Autocomplete plugin wrapper
+ @author Anatoly Mironov
+ http://sharepointkunskap.wordpress.com, http://www.bool.se
+
+ @param [ui-autocomplete] {object} Options to pass to $.fn.autocomplete() 
+ */
+
+angular.module('ui.directives')
+
+.directive('uiAutocomplete', [function() {
+  var directive = {
+		require: '?ngModel',
+		link: function(scope, element, attrs, controller) {
+			var getOptions = function() {
+				return angular.extend({}, scope.$eval(attrs.uiAutocomplete));
+			};
+			var initAutocompleteWidget = function () {
+				var opts = getOptions();
+				element.autocomplete(opts);
+				if (opts._renderItem) {
+					element.data("autocomplete")._renderItem = opts._renderItem;
+				}
+			};
+			// Watch for changes to the directives options
+			scope.$watch(getOptions, initAutocompleteWidget, true);
+		}
+	};
+  return directive;
+}]);
--- a/modules/directives/autocomplete/autocomplete.js
+++ b/modules/directives/autocomplete/autocomplete.js
@@ -9,22 +9,23 @@
 
 angular.module('ui.directives')
 
-.directive('uiAutocomplete', [function() {
+.directive('uiAutocomplete', ['ui.config', function(uiConfig) {
   var directive = {
 		require: '?ngModel',
 		link: function(scope, element, attrs, controller) {
-			var getOptions = function() {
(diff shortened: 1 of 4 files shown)


modules/directives/autocomplete/autocomplete.js
```diff
@@ -0,0 +1,31 @@
...
+ jQuery UI Autocomplete plugin wrapper
+ @author Anatoly Mironov
+ http://sharepointkunskap.wordpress.com, http://www.bool.se
+
+ @param [ui-autocomplete] {object} Options to pass to $.fn.autocomplete() 
+ */
+
+angular.module('ui.directives')
+
+.directive('uiAutocomplete', [function() {
```
ProLoser: Can you add support for uiConfig?

modules/directives/autocomplete/autocomplete.js
```diff
@@ -0,0 +1,31 @@
...
+			};
+			var initAutocompleteWidget = function () {
+				var opts = getOptions();
+				element.autocomplete(opts);
+				if (opts._renderItem) {
+					element.data("autocomplete")._renderItem = opts._renderItem;
+				}
+			};
+			// Watch for changes to the directives options
+			scope.$watch(getOptions, initAutocompleteWidget, true);
```
ProLoser: This is evaluating the options every single `$apply()` which is very inefficient.

Instead, I would just initialize the plugin normally once and then do something like:

``` js
scope.$watch(attrs.uiAutocomplete, function(newVal, oldVal){
  element.autocomplete('options', newVal);
}, true);
```

modules/directives/autocomplete/autocomplete.js
```diff
@@ -0,0 +1,31 @@
...
+		require: '?ngModel',
+		link: function(scope, element, attrs, controller) {
+			var getOptions = function() {
+				return angular.extend({}, scope.$eval(attrs.uiAutocomplete));
+			};
+			var initAutocompleteWidget = function () {
+				var opts = getOptions();
+				element.autocomplete(opts);
+				if (opts._renderItem) {
+					element.data("autocomplete")._renderItem = opts._renderItem;
```
ProLoser, line 24: What is the purpose of `_renderItem`?

mirontoli, line 24: _renderItem is a "private" scoped function in jQuery UI autocomplete which defines how an item has to be rendered. Otherwise a plain list is rendered. I wrote about it in [my blog post](http://sharepointkunskap.wordpress.com/2013/01/04/angular-jquery-ui-autocomplete/). Previously `formatItem` was an option in autocomplete, but no longer.

modules/directives/autocomplete/autocomplete.js
mirontoli: Sure, I'll try to find out how to do it, and will do it. Could you provide a link where uiConfig is explained?

modules/directives/autocomplete/autocomplete.js
mirontoli: Great comment. I aggree. This is a little bit inefficient. I'll change..

modules/directives/autocomplete/autocomplete.js
ProLoser: Poke around some other directives and see how we did it. For example uiJq, uiSelect2, uiDate, etc. It's fairly simple.

mirontoli: Now I understand the purpose of [uiConfig](https://github.com/angular-ui/angular-ui/tree/master/modules/directives/jq) It is really awesome!

PR #419, opened 2013-02-12
feat(debounce): Add initial service factory - NEEDS TESTS and DOCS

ProLoser:
This PR is to remind me to follow up on this.


petebacondarwin:
Here is a more correct demo http://plnkr.co/edit/afX9v0?p=preview

ProLoser:
I'm closing this as I don't think it adds anything angular-specific over using something like _.js

PR #424, opened 2013-02-13
Added keep-masked option.

carlosbrando:
Sometimes we need to get the masked value of an input, such as a date "10/12/2012".

In this case just add data-keep-mask="true" in the input.

### DIFF
--- a/modules/directives/mask/mask.js
+++ b/modules/directives/mask/mask.js
@@ -29,7 +29,11 @@ angular.module('ui.directives').directive('uiMask', [
          */
         element.bind('keyup', function () {
           $scope.$apply(function () {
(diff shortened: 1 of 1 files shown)


ProLoser:
The mask plugin doesn't take this as an option natively?

Also, shouldn't you just avoid all the code in the bind altogether?

Also, you should probably do `attrs.keepMasked`

Also, we may want to use a more appropriate property name (or make it a global config/passed option?)

shaungrady:
The uiMask directive has since been refactored from the ground-up—this commit is no longer applicable.

In the new directive, the masked value can easily be read from ngModel controller's $viewValue attribute of the input and is updated with every keystroke.

PR #438, opened 2013-02-17
Select 2 bug fixing

atruskie:
As part of being a good citizen, a directive that uses NgModel, should (AFAIK)
use the correct values to render elements. Previously the select control
was using modelValue as opposed to viewValue inside the $render method.
This meant that cool tricks (like injecting parsers and formatters did
not work properly).

Also changed the directive initialisation to be more robust. In certain
cases the initial value was not being set correctly... meaning that on
recreation, any bound data would not show up properly.

This is my first pull request (ever). I have written tests for all the work I have done. Let me know if I did anything wrong :-)

### DIFF
--- a/modules/directives/date/test/dateSpec.js
+++ b/modules/directives/date/test/dateSpec.js
@@ -1,4 +1,4 @@
-/*global describe, beforeEach, it, inject, expect, module, $*/
+/*global describe, beforeEach, afterEach, it, inject, expect, module, dump, $*/
 describe('uiDate', function() {
   'use strict';
   var selectDate;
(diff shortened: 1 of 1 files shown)


ProLoser:
Can you remove your build files from the commit? Just commit the affected source files. You can --amend and then force-push the new version of the same commit.

You should also open pull requests from a branch on your fork, rather than from your master. 

## 

Dean Sofer
DeanSofer.com
714.900.2254

atruskie:
I _think_ I did what you asked...?

ProLoser:
No you deleted all the build files from the repo. You are supposed to simply 'not build'. You need to uncommit your changes to the build files. Under 'Files Changed' it should say only 2: the source file and the test file.

atruskie:
I've made the correct changes to a different fork, should I close this pull request and reopen a new one?

ProLoser:
Yeah

Issue #433, opened 2013-02-15
Sortable: values not always updated


zwack:
When using the sortable demo the UI is working fine all the time. However the values in the JSON structure are not always updated. I couldn't reproduce how to reliable trigger the bug, it seems to happen more often when moving items top down though.

I got the same behaviour using the sortable directive locally. Tested in Linux/Windows on Chrome 24 and FF 18.

Thanks for your work!

Issue #394, opened 2013-02-01
Problems with item being duplicated when using ui-sortable with connected nested lists


aravinda0:
I've been trying to use this commit for ui-sortable: 
https://github.com/angular-ui/angular-ui/commit/0d86357afaa3da22ffdc244c892112c640a85802

Take this plunker: http://plnkr.co/edit/58O5JW?p=preview
When you move the item titled 'Samsung to release...' into the list titled 'Top News', the item gets duplicated.

Using a newer version of jquery-ui (I used 1.10.0) solves that issue. 

But just upgrading to a newer jquery-ui version doesn't fix everything.
http://plnkr.co/edit/lSErMO?p=preview

In the above plunker, if you move the 'New Phone Releases' sublist to the END of the list titled 'Top News', the move takes place, but an empty element also gets appended to the latter.

I'm quite new to javascript and angular, but the following crude modification in the ui-sortable directive seems to address the above issue. The changes are in the sortable.js file.
http://plnkr.co/edit/VKjbhV?p=preview

Please review the code.

ProLoser:
You're using `connectWith: '.target'` but `class="target"` is on both the parents AND the children. You must use a separate class or selector to differentiate between the two levels.

I feel like I have de-ja-vu. I know I answered this exact question at least once before.

ProLoser:
You saved over the middle link when you posted the third link (same url). I am unable to see a 'samsung' item in all of your links provided.

aravinda0:
Damn. Sorry about the duplicate link. I've updated the middle link.

The 'samsung..' item is in the first plnkr link. The list is a bit long due to margins and all, so if you're on a smaller screen, try clicking on 'launch preview in a separate window' button at the top right of the preview section.

And yes, the deja-vu is because you told me on IRC about the reason that there is the same class on both the parent list and child list.
I'm trying to make modifications to the code to allow items in the parent list to be moved into a child list.

ProLoser:
I think I was noticing this bug too when I was putting the demo together. I haven't been able to thoroughly review the new source for the error, but I will try to do this before v0.4.0

If you want to help me out, you can try inspecting the source code yourself and help me out by helping me identify the problem code. Keep in mind all that is really being done is that the index of the li you are dragging is mapped to the ng-model array, and when you drop, the source array gets the item spliced out and the target array gets the item spliced in.

aravinda0:
[...]

PR #442, opened 2013-02-18
Fix synchronization between sorted list and model

skahlert:
I'm not sure what's the reasoning behind those two lines, but removing them seems to fix Issue #433.

### DIFF
--- a/modules/directives/sortable/sortable.js
+++ b/modules/directives/sortable/sortable.js
@@ -51,8 +51,6 @@ angular.module('ui.directives').directive('uiSortable', [
               var end, start;
               start = ui.item.sortable.index;
               end = ui.item.index();
-              //if (start < end)
-              //  end--;
 
               // Reorder array and apply change to scope
               ui.item.sortable.resort.$modelValue.splice(end, 0, ui.item.sortable.resort.$modelValue.splice(start, 1)[0]);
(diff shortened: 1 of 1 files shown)


ProLoser:
Need testing and some more information as to why this fixes it. Any plunkers?

skahlert:
Well, here's one:
http://embed.plnkr.co/ldSAFqHq4Gdbsvb24n2i
But behavior is the same on the demo page (see: https://github.com/angular-ui/angular-ui.github.com/issues/56)
Refresh each time:
- drag Barney to top position -> model updated correctly
- drag Anna to second position -> Anna in model stays in top position
- drag Anna to third position -> Anna in model moves to second position

Its almost like the lines were put there to create exactly that behavior, but it isn't quite how you'd expect it to behave.

Here's one with the fix:
http://embed.plnkr.co/A26vnnQi6wd1uzXJ7JLx

ProLoser:
Can you compare your version against #394

ProLoser:
If you test all the behaviors he mentions in the beginning of the ticket and you can confirm that this version does not recreate the problems described then I will make the necessary changes.

The main concern is when you work with lists that are not connected. I figured that the specific line in question could have been a problem though.

skahlert:
Well... Yeah... I think it works for issue #394, too.

At least with the the example containing the newer jQueryUI version.
http://plnkr.co/edit/1ybNAQ

The first example seems broken with or without the pull-request, though.

ProLoser:
So it was a jQueryUI bug?

## 

Dean Sofer
DeanSofer.com
714.900.2254

ProLoser:
Cool well can you change your PR to delete the lines instead of commenting them out?

## 

Dean Sofer
DeanSofer.com
714.900.2254

skahlert:
No worries... already done.

PR #446, opened 2013-02-19
fix(mask): mask can now work with original jquery.maskedinput plugin

just-boris:
I move mask value validation into the directive and update jquery.maskedinput plugin to lastest version. Plugin now compatible with jquery 1.9.
Furthermore I add test of optional part of mask, because it is important part of my changes


ProLoser:
Can you go through this list: https://github.com/angular-ui/angular-ui/blob/master/CONTRIBUTING.md

just-boris:
Sorry for first commit, now that looks better

just-boris:
I found issue in Opera in this fix. I need make some changes before committing it again

just-boris:
Sometimes the value is set from element.val(). Now parser always gets value from maskedinput directly and ignores other values.
Moreover, current master has the error: http://screencast.com/t/yBGSgfFcL0ip (Opera 12.14, IE9)
After this fix it works normal

shaungrady:
uiMask has since been refactored, this issue is no long applicable. It will be in the next AngularUI release. jQuery dependency has been removed.

PR #451, opened 2013-02-19
Poposed fix for issue #450

### LINKED ISSUE #450: Sortable clobbers scope of other sortable instances
It looks like options passed to ui-sortable are overwritten by later instances of sortable. Especially the callbacks (stop for example) seem to reference the wrong scope, when called.
Here's an an example:
http://plnkr.co/edit/o4Sgr6NW0FJrYoigbFOX
Notice, no matter what you sort, it will alert the lower list.

skahlert:
This should fix #450 
Example:
http://plnkr.co/edit/r7iaAsIGO3j5VwjVux9k

### DIFF
--- a/modules/directives/sortable/sortable.js
+++ b/modules/directives/sortable/sortable.js
@@ -8,7 +8,7 @@ angular.module('ui.directives').directive('uiSortable', [
     return {
       require: '?ngModel',
       link: function(scope, element, attrs, ngModel) {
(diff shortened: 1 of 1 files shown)


skahlert:
You're right... that one cost me a lot of sleep yesterday, and I finally assumed it was some weird way how angular works.
It turns out, the `_stop` variable just wasn't declared in the link function. Either that variable leaked in somewhere from the outside, or it was otherwise just treated as a global variable...
Declaring it should make the closures redundant.

PR #457, opened 2013-02-23
fix(select2): Updating Select2 version support

ProLoser:
No idea why the tests are failing. When `allowClear:true` setting the val to falsey values should work fine.


ProLoser:
Appears to be present in the latest version of angular-ui/ui-select2

dlidstrom:
When can we expect to see this fix in a release version? 0.4.0 seems to have the bug.

PR #467, opened 2013-02-27
Fix UI refresh on programmatic change

zupolgec:
On model variable change, the UI doesn't get the change because the trigger call is missing.

### DIFF
--- a/modules/directives/select2/select2.js
+++ b/modules/directives/select2/select2.js
@@ -65,6 +65,10 @@ angular.module('ui.directives').directive('uiSelect2', ['ui.config', '$timeout',
                 }
               }
             }
+
+            $timeout(function () {
+              elm.trigger('change');
+            });
           };
 
           // Watch the options dataset for changes


ProLoser:
What does this fix exactly since you're setting the value using `elm.select2('val')` or `('data')`?

There is also no explanation as to why it's wrapped in a timeout.

zupolgec:
Triggering the change event fixes the asynchrony of the <select> element and the select2 UI.

It's wrapped in a timeout (like 10 rows below) because the change event that get's triggered (20 rows below) uses the scope.$apply function that raises an error if called during $digest and with $timeout you avoid that.

ProLoser:
So instead of delaying things and fixing symptoms which creates new bugs and requires more symptom avoidance, why not fix the problem? You can check the `$$phase` and avoid calling $apply() instead.  

Anyway, I still don't really know why you need to trigger a change. Select2 should update if you use their methods to set the value. Can you explain specifically why triggering a change event helps or addresses?  

## 

Dean Sofer
DeanSofer.com
714.900.2254

zupolgec:
I don't know why you are criticizing this solution (the $timeout) when it's already used in this file.

Maybe the issue I have is caused by using `ng-options` that is not suggested, but I can't avoid.

ProLoser:
There is a concrete reason $timeout is being used, and it is not to avoid a $scope.$apply(). The reason a $timeout is used is so that the child options have time to render and Select2 can see those generated DOM elements. I'm not trying to be purely critical, you should go checkout $$phase when working with angular.

If you do not follow the directions and experience problems, you can't really submit patches, as you're fixing symptoms of a broken edge-case.

What is the reason that you must use ng-options?

On a side note, I've been talking to another developer about finally taking another crack at rewriting ui-select2 to go back to use ng-options.  

## 

Dean Sofer
DeanSofer.com
714.900.2254

zupolgec:
I use ng-options because I can do something like `ng-options="car.model group by car.make for car in cars"` and have the options grouped. Is there another way?

I can't make this work without $timeout because I keep getting error about digest already in progress.

ProLoser:
Do you mind relocating this to https://github.com/angular-ui/ui-select2

PR #472, opened 2013-03-03
Fix broken ui-validate-watch for retroactive dependencies

ovax3:
In the `ui-angular-watch` implementation we should not re-validate against the `$modelValue` but the `$viewValue` of the field since : 
1. this is what angular naturally does ;
2. most of the time the `$modelValue` is `undefined` since angular values it if and only if the `$viewValue` is valid first.

The problem can be reproduced on the demo page by filling the `passwordConfirmation` first, and then the `password`. The `passwordConfirmation` stays invalid which is not expected.

The attached pull request fixes the problem by triggering again the natural angular validation process by simulating a change in the `$viewValue`

```
ctrl.$setViewValue(ctrl.$viewValue);
```

For instance, with this fix, the following cyclic five fields dependency schema works like a charm

```
A -> B -> C -> D -> E -> A
```

### DIFF
--- a/component.json
+++ b/component.json
@@ -2,7 +2,7 @@
   "author": "AngularUI Team",
   "name": "angular-ui",
   "description": "AngularUI - Companion Suite for AngularJS",
-  "version": "0.4.0",
+  "version": "0.4.1",
   "homepage": "http://angular-ui.github.com",
   "repository": {
     "type": "git",
--- a/modules/directives/validate/test/validateSpec.js
+++ b/modules/directives/validate/test/validateSpec.js
@@ -1,5 +1,5 @@
 describe('uiValidate', function ($compile) {
-  var scope, compileAndDigest;
+  var scope, compileAndDigest, changeInputValue;
 
   var trueValidator = function () {
     return true;
@@ -25,6 +25,10 @@ describe('uiValidate', function ($compile) {
       scope.$digest();
 
       return inputElm;
+    };
+    changeInputValue = function(elm, value) {
+      elm.val(value);
+      elm.trigger('input');
     };
   }));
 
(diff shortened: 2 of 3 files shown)


ProLoser:
Since you are triggering the entire stack of validators there is no reason to have multiple watches, and since you have multiple watches there is no longer a reason to wrap the expression in a string and double evaluate it. 

You should also update the tests and possibly add one for your patch

ovax3:
String removed. Much simpler. Not backward compatible.
Tests updated.
Please ignore the commit `8c084ea: Fake version tag`. Nothing to do with this issue.

ovax3:
Any news about this issue ?

inklesspen:
I ran into this same issue today. Is there a temporary workaround while we wait for the pull request?

ProLoser:
The pull request IS the workaround.

rintaun:
This works great, thanks! One thing worth mentioning is that the model in the `ui-validate-watch` attribute must not be quoted as currently shown in the docs.

And one problem -- this method sets the field to dirty right off the bat, which can be problematic for styling.

dmackerman:
This isn't working for me in 1.2. I'm trying to verify that a specific field is less than another, and vice versa.

``` html
<input type="text" autocomplete="off" id="startDate" maxlength="4" name="startDate" 
  ng-model="education.startDate" 
  ui-validate="$value <= education.endDate"
  ui-validate-watch="education.endDate" required />

<input type="text" autocomplete="off" id="endDate" maxlength="4" name="endDate"
  ng-model="education.endDate" 
  ui-validate="$value >= education.startDate"
  ui-validate-watch="education.startDate" required />
```

Am I doing something wrong here?

PR #479, opened 2013-03-10
Mask refactor

shaungrady:
No more dependencies. Not even jQuery.
I've never written an end-to-end test before, so I'll need assistance rewriting the spec.

Big things to note about the refactor:
- $observe is set on the mask attr, so mask is fully dynamic.
- Model is now sent the unmasked value of the input. This is more useful, arguably, than the masked value for insertion into a database. Masking is a function of the view, not the model.

This is my first real directive I've written, so I'd love all the feedback I can get.
Thanks.

### DIFF
--- a/modules/directives/mask/demo/index.html
+++ b/modules/directives/mask/demo/index.html
@@ -5,8 +5,24 @@
       angular.module('ui.directives',[]);
     </script>
     <script type="text/javascript" src="../mask.js"></script>
+    <style type="text/css">
+      input, button { font-size: 14px; font-family: helvetica, arial; width:300px; text-align: left }
+    </style>
   </head>
   <body ng-app="ui.directives">
-    <input ui-mask="'(9)9'" ng-model="x">
+
+    <label>
+      Masked Input:<br>
+      <input ui-mask="{{y}}" ng-model="x" placeholder="Specify a mask or click a button">
+    </label><br>
+    <label>
+      Mask Defintion:<br>
+      <input ng-model="y">
+    </label><br><br>
+
+    <button ng-click="y = undefined">Set mask to undefined (uninitialize)</button><br>
+    <button ng-click="y = '999-9999'">Set mask to 999-9999</button><br>
+    <button ng-click="y = '(999) 999-9999'">Set mask to (999) 999-9999</button><br>
+    <button ng-click="y = '(999) 999-9999 ext. 999'">Set mask to (999) 999-9999 ext. 999</button>
   </body>
 </html>
--- a/modules/directives/mask/demo/index.html
+++ b/modules/directives/mask/demo/index.html
@@ -1,0 +1,12 @@
+<html>
+  <head>
+    <script type="text/javascript" src="http://code.angularjs.org/1.1.3/angular.js"></script>
+    <script type="text/javascript">
+      angular.module('ui.directives',[]);
+    </script>
+    <script type="text/javascript" src="../mask.js"></script>
+  </head>
+  <body ng-app="ui.directives">
+    <input ui-mask="'(9)9'" ng-model="x">
+  </body>
+</html>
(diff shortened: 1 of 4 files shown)


petebacondarwin:
Hi Shaun
This is excellent stuff!
I pulled your branch and made some changes:
https://github.com/petebacondarwin/angular-ui/tree/mask
See what you think.  Still not sure how to test UI triggered changes.
Pete

petebacondarwin:
By the way, I think that dynamic masks should be done with interpolation,
which removes the need to have single quote around every static mask:

ui-mask="{{someMaskVar}}"

shaungrady:
Hi Pete, 

Thanks! I agree, interpolation does make more sense. I will work on that today. I will also work on making some of the comments in the code more helpful. 

I'm not sure how to test UI-triggered events either. What do you think of this approach: 

On line 93, where the event listener is attached, adding a custom event…

```
iElement.bind('input propertychange keyup click mouseout maskdebug', eventHandler);
```

… and add a second argument to the eventHandler func for testing, have it do something like…

```
function eventHandler(e, eDebug) {
    var isDebug = (e.type == 'maskdebug') && angular.isObject(eDebug);
    e = isDebug ? eDebug : e;

    ...

    // At the end of the handler, hit the testing callback
    if (isDebug)
        e.callback(valMasked, caretPos);
```

Then, in the test, we can use jQuery's .triggerHandler() and pass our own hand-made event object to the handler.

```
function testCallback(valMasked, caretPos) {
    // test to see if caret moved leftwards passing over all mask-component chars
}
element.triggerHandler('maskdebug', { which:37, callback:testCallback });
```

What do you think?

petebacondarwin:
RE triggering - Maybe ...
I have changed the code on my branch to use interpolation already.  See the
unit tests.

shaungrady:
Ah I see, fantastic! Other than tests, is there anything else that needs to be done?

petebacondarwin:
Documentation - particular updating the gh pages index.html!
Also I was unsure about the input being cleared when the model is not
valid?  Shouldn't it look like it does when the masked input is first
rendered?
And also, I notice that you only update the model on blur.  This is not in
line with how other AngularJS input directives work.  Is there some reason
why it shouldn't be on every change?

shaungrady:
Input clearing on render:
I have mixed feelings about this. Even if the model is considered invalid by the mask directive on render, it seems wrong to have the a directive change the model without user interaction. The form would immediately be marked dirty and, if ngRequired, marked invalid. It feels more appropriate to leave the model untouched, even if it doesn't jive with the specified mask, until a user touches it. What are your thoughts on this?

The model is updated through controller.$parsers on every user input. It's not updated on blur, though, which is why I call $setViewValue to update the model to the cleared state (if mask was invalid when user blurred).

shaungrady:
```
This is not in line with how other AngularJS input directives work
```

Just to expand on my response on this, I believe it's the ngModel directive that's already pushing the input's value to the model; the mask directive needs only to add it's own formatter and parser. The $setViewValue in blurHandler makes sure the .val('') clearing on invalid mask goes through the parsing pipeline too. Is there a more appropriate way to handle this?

petebacondarwin:
This is all looking excellent now!  I think it is working as I would expect.

petebacondarwin:
Merged into master.

petebacondarwin:
Thanks Shaun for this cracking PR

shaungrady:
Thanks for the assistance Pete. There is another "hidden" feature of the mask where it can accept an array, allowing you to define your own RegEx patterns, as shown here: http://plnkr.co/edit/qKCgHmJqOg7iXczmup5i?p=preview

Though I'd like to think about it's implementation a bit more before really documenting it.

courtyenn:
Would anyone know if we can use this directive in the angular-ui-mask repository? I would like to send back the formatted value back to the server, and in the documentation, it isn't clear. Thanks!

PR #482, opened 2013-03-11
Can register characters for keypress command, not just codes

dturkenk:
Can register an actual character in the keypress, keyup and keydown
directives (e.g. 'a' instead of 65/97)

Handles the different values for keycodes on keyup/down and keypress
both with and without JQuery.

Tested with automated tests, IE9, latest Chrome and Firefox on Windows

**Important:** characters will be converted to lower-case. If you want to capture an upper-case character, use "shift-char"


ProLoser:
Do you mind relocating this to https://github.com/angular-ui/ui-utils ?

dturkenk:
Will do. I'll move it it over there in the next few days.

Thanks.

dturkenk:
Partially ported this to ui-utils project.

Closing the pull request here.

PR #486, opened 2013-03-15
fix(keypress): allow events to ommit key modifiers

caiotoon:
Programmatically triggered events that do not include key modifiers (`altKey`, `shiftKey`, `ctrlKey`) properties would fail even when combination didn't require these keys.

### DIFF
--- a/modules/directives/keypress/keypress.js
+++ b/modules/directives/keypress/keypress.js
@@ -45,9 +45,9 @@ angular.module('ui.directives').factory('keypressHelper', ['$parse', function ke
     // Check only matching of pressed keys one of the conditions
     elm.bind(mode, function (event) {
       // No need to do that inside the cycle
-      var altPressed = event.metaKey || event.altKey;
-      var ctrlPressed = event.ctrlKey;
-      var shiftPressed = event.shiftKey;
+      var altPressed = event.metaKey || event.altKey || false;
+      var ctrlPressed = event.ctrlKey || false;
+      var shiftPressed = event.shiftKey || false;
       var keyCode = event.keyCode;
 
       // normalize keycodes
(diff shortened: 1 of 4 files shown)


ProLoser:
Strange, no idea why we're even using this `|| false` pattern when we should probably just be using `!!`

ProLoser:
How is your tests any different from https://github.com/caiotoon/angular-ui/blob/7a9dbcf4d9dd9a83606e51b8121437a82d52d48a/modules/directives/keypress/test/keypressSpec.js#L32-L35
?

I don't see the purpose of this.

caiotoon:
You're right regarding `|| false` pattern. `!!` is better. Updated, also changed for other parts of code previously committed this way.

Regarding the test case, the only difference is that `event.altKey`, `event.ctrlKey` and `event.shiftKey` are not defined, leading the handlers to never run.

As you can see, in my test cases I don't use the `createEvent` utility. Do you believe we could remove these test cases due to its simplicity?

caiotoon:
This is the test that would fail:

``` javascript
it('should support modifiers to be ommited from event', function () {
  var keyEvent = jQuery.Event('keydown');
  keyEvent.keyCode = 13;
  createElement({'13': 'event=true'}).trigger(keyEvent);

  expect($scope.event).toBe(true);
});
```

Notice that `keyEvent` is not created by utility and the modifiers key are not set.

ProLoser:
If you would like to refactor the tests in this PR, go ahead. It may be more relevant to remove your test too then since that would mean the very first test and yours would then be identical.

caiotoon:
Just did it. Test utility do not send key modifiers unless explicitly defined. This way will cover all cases, and make my tests unnecessary indeed.

Issue #434, opened 2013-02-15
route directive fails to create scope


tisztamo:
It seems to me that the uiRoute directive doesn't create a new scope so if I use it on multiple tags the last $uiRoute wins.

This is not working:

``` html
<li data-ui-route="/public" ng-class="{active: $uiRoute}"><a href="#">Public</a></li>
<li data-ui-route="/private" ng-class="{active: $uiRoute}"><a href="#">Private</a></li>
```

But this is OK:

``` html
<li ng-repeat="btn in navbuttons" ui-route="{{btn.route}}" ng-class="{active: $uiRoute}"><a href="#">{{btn.label}}</a></li>
```

(The demo on the site uses ng-repeat which creates a new scope for every iterated tag so it is working correctly.)

sudhakar:
I had the same problem too. After setting `scope: true` on the uiRoute solves the problem as new scope is created for it.

ProLoser:
Man I never thought about that. Thanks for the tip, will change immediately.

ProLoser:
Crap, how the fuck do you test child scopes.

ProLoser:
If anyone wants to help fix the tests with `scope:true` it would help me out a lot

ajoslin:
Did you try something like this?

``` js
var elm = $compile('<my-directive></my-directive>');
var childScope = elm.children().scope();
```

ProLoser:
Is that the proper way to do it? I mean the element itself has the scope. Or maybe I should just do `elm.scope()`. I just sort of figured doing `scope()` was a hackish way to go about things.

ProLoser:
Also, the dom element doesn't have children.

leeola:
@sudhakar re:

> I had the same problem too. After setting scope: true on the uiRoute solves the problem as new scope is created for it.

How exactly would you do that?

I'm not familiar with Angular/AngularUI's method for providing options to modules, but when i tried to emulate the `ui-select2` example shown on the homepage, i couldn't seem to set `scope: true`.

I tried

```
myModule.value('ui.config', {
  uiRoute: {
    scope: true
  }
});
```

with the names `uiRoute`, `route`, and `'ui-route'` to no avail. Any advice?

sudhakar:
Nope. For the time being, I have modified the uiRoute directive itself with `scope: true`, so that it creates a new scope for the element & will be available for its child DOM as well. As of now, it works for me, but I havent tested it in other scenarios.

[...]

PR #489, opened 2013-03-17
Fix for #434

ataube:
Hey @ProLoser: 
I finally found a solution to fix the tests with scope = true for the ui-route directive. Note the multi scope support is not tested yet. I made some experiments with the following example which works fine if there is no ng-model defined but I was not able to integrate it without a big refactoring of the routeSpec...

<pre><code>
it('should support multiple scopes', function() {
      setPath('/foo');
      elm = $('&lt;ui&gt;&lt;li ui-route=&quot;/foo&quot;&gt;Foo&lt;/li&gt;&lt;li ui-route=&quot;/bar&quot;&gt;Bar&lt;/li&gt;&lt;/ul&gt;');
      $compile(elm[0])(scope);
      expect(elm.find('li').eq(0).scope()["modelProp"]).toBe(true);
      expect(elm.find('li').eq(1).scope()["modelProp"]).toBe(false);
      setPath('/bar');
      expect(elm.find('li').eq(0).scope()["modelProp"]).toBe(false);
      expect(elm.find('li').eq(1).scope()["modelProp"]).toBe(true);
    });
</code></pre>

### DIFF
--- a/modules/directives/route/route.js
+++ b/modules/directives/route/route.js
@@ -4,6 +4,7 @@
 angular.module('ui.directives').directive('uiRoute', ['$location', '$parse', function ($location, $parse) {
   return {
     restrict: 'AC',
+    scope: true,
     compile: function(tElement, tAttrs) {
       var useProperty;
       if (tAttrs.uiRoute) {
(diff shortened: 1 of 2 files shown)


ProLoser:
Thanks. I hate when broken tests block a perfectly viable patch.

PR #514, opened 2013-03-26
Angular-UI autocomplete wrapper

JakobJingleheimer:
Hello,

I've created a wrapper for the jQueryUI autocomplete widget (based on uiDate). ~~This does have a known bug where selecting a suggestion via mouseclick does not cause the model to update (but using arrow keys + return does update the model). jQueryUI's autocomplete doesn't have an equivalent `.getDate()` method to get the selected item, so I used `element.val()`, which does not seem to update when the select event is fired from mouseclick.~~ (fixed in 92778051028800958587282b4671982e5b0ac6bd)

This directive supports uiConfig and extends the defaults with a scope object passed from uiAutocomplete.

I saw #414, but the code didn't feel like it came from the AngularUI, and the way it required functions in the controller did not seem to be the "Angular Way" of doing things. Also, I couldn't get it to work.

P.S. Sorry if I've done this pull request incorrectly—I'm new to Github and this is my first PR.

### DIFF
--- a/modules/directives/autocomplete/autocomplete.js
+++ b/modules/directives/autocomplete/autocomplete.js
@@ -21,15 +21,12 @@ angular.module('ui.directives')
         require: '?ngModel',
         link: function (scope, element, attrs, controller)
         {
-          var getOptions = function ()
-          {
-            var options = angular.extend({}, uiConfig.autocomplete, scope.$eval(attrs.uiAutocomplete));
-            return options;
-          };
 
           var initAutocomplete = function ()
           {
-            var opts = getOptions();
+            var opts = angular.extend(
+              {}, uiConfig.autocomplete, scope.$eval(attrs.uiAutocomplete)
+            );
 
             // If we have a controller (i.e. ngModelController) then wire it up
             if ( controller )
@@ -53,7 +50,7 @@ angular.module('ui.directives')
             element.autocomplete(opts);
           };
           // Watch for changes to the directives options
-          scope.$watch(getOptions, initAutocomplete, true);
+          scope.$watch(attrs.icAutocomplete, initAutocomplete, true);
         }
       };
     }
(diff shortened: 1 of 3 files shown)


JakobJingleheimer:
Why does uiDate [watch `getOptions`](https://github.com/angular-ui/angular-ui/blob/master/modules/directives/date/date.js#L72) instead of `attrs.uiDate`?

thyco:
Hey, was just trying to use the autocomplete directive, but it doesn't seem to handle prefilled model values very well.
This plunker shows what I mean : http://plnkr.co/edit/BUDxTx7rB2UlqtmcSvlg?p=preview

I would expect the input field to show 'item 1', not [object Object]. Any ideas on how to fix this?

JakobJingleheimer:
@thyco, I believe _ngModel_ expects a string: http://docs.angularjs.org/api/ng.directive:input.text

When I did this for my own project, my initial value supplied to _ngModel_ was a string, so I did not encounter your issue. However, the _uiAutocomplete_ directive replaces the value of the input field with the `ui.item.value` property; _ngModel_ receives the whole `ui.item` object.

thyco:
I took a further look at it and came up with something that seems to work.

http://plnkr.co/edit/ncEWOuMrMTJT9kK4n5s0?p=preview

Basically I have overwritten the controller.$render function to display the value instead of the object itself. The code should still work with only strings (not sure, haven't tested it). The code does assume that the objects in the list have a value property.

JakobJingleheimer:
@thyco, I believe it is safe to assume that there will be a value property returned in the jQueryUI object: If `source` is a list of strings, jQuery uses that string to define both the `ui.item.value` and `ui.item.label`

However, if, in the object you initially pass to _ngModel_ a `value` property is not defined, it still displays `[object Object]`. The initial value of _ngModel_ is not used by jQueryUI/uiAutocomplete so "we" would be imposing the requirement that the initial object contain a `value` property.

Forked plunkr to demonstrate both: http://plnkr.co/edit/WBltgj

wesleycho:
This repository is discontinued - this might be better relocated as a separate module from `ui-utils`, the current core version of this repository.

I recommend opening an issue, and having discussion continue there on this feature.

PR #515, opened 2013-03-27
Digest on event change originating from Calendar

glebm:
This pull requests triggers view updates when the user drags or resizes events from FullCalendar

### DIFF
--- a/modules/directives/calendar/calendar.js
+++ b/modules/directives/calendar/calendar.js
@@ -45,7 +45,13 @@ angular.module('ui.directives').directive('uiCalendar',['ui.config', '$parse', f
               var expression,
                 options = {
                   defaultView : view,
-                  eventSources: sources
+                  eventSources: sources,
+                  eventDrop: function() {
+                    scope.$apply();
+                  },
+                  eventResize: function() {
+                    scope.$apply();
+                  }
                 };
               if (attrs.uiCalendar) {
                 expression = scope.$eval(attrs.uiCalendar);


joshkurz:
I see what you trying to do and thanks for the PR.

I have a few questions however?

What if someone wanted to do more than just update the view? 

Does this hardcode in the eventDrop and resize method to where they cannot be extended from the controller and passed in? 

If they extend the eventDrop method from the controller then do they still need to manually call apply or would this cure that? 

There needs to be some tests written.

glebm:
While I do think it should be possible to bind directly to fullCalendar events, the basic behaviour of keeping view and the model in sync should always be there, because that's how all angular components behave as far as I know

> There needs to be some tests written.

I am new to angular, haven't looked into writing tests

By the way, the default equality check doesn't take contents into account.
We could just convert the whole array to json (fast because it's native) and compare the strings to detect everything

joshkurz:
Not a bad idea. I tried to do something like this in the beginning, but ran into problems with digest iterations. 

The browser would crash and burn if we tried to do a deep watch on the event objects. I think that's what your referring to as the default equality check. 

So the workaround is to keep the calendar in sync from the controller on each change. This is just a workaround however so if we can get the calendar to efficiently check the contents of the events for changes rather than just the length of each event source, then we would be in the right track. 

The goal is to make sure its efficient while being purely self sufficient.

glebm:
I've attempted it here https://github.com/glebm/angular-ui/blob/calendar/modules/directives/calendar/calendar.js, and so far it works on the example data. 
Instead of doing toJson in the end I just do Array.prototype.join on specific attributes. This should scale, but I haven't tried.

glebm:
JSPerf! http://jsperf.com/angular-calendar-events-fingerprint

joshkurz:
Hmm very nice. We can play with this method for sure. Thanks

Since I am in the middle of moving the uiCalendar to the new repo this PR will not be merged in, but once we move the calendar over we can make a new PR on that repo with these changes and all any others desired.

Again thanks for the help on this as I believe that you are spot on.

glebm:
Good news! I managed to get some ridiculous performance ( >1000 events fingerprinted and compared over 1,500 times per second) by doing a simple trick here http://jsperf.com/angular-calendar-events-fingerprint/3 (last test case)

Turns out if we convert dates to int with a `+`, before converting `toString()`, we get 8x performance gain in Chrome and 20x performance gain in Firefox. 
In practice this means one can have tens of thousands of events in the calendar without noticing any responsiveness loss (and FullCalendar is awesome because it only processes events in the current view's date range).  This only unknown variable is how Internet Explorer deals with it (I will check IE9 soon), as usual :) 

> Again thanks for the help on this as I believe that you are spot on.
> I'm using angular-ui in a commercial app (i.e. to make money), thanks for building this in the first place!!

joshkurz:
This looks really spectacular. I think that this is $. 

Glad someone else is making cheddar off of AngularJS. :)

On Fri, Mar 29, 2013 at 10:05 AM, Gleb Mazovetskiy <notifications@github.com

> wrote:
> 
> Good news! I managed to get some ridiculous performance ( >1000 events
> fingerprinted and compared over 1,500 times per second) by doing a simple
> trick here http://jsperf.com/angular-calendar-events-fingerprint/3 (last
> test case)
> 
> Turns out if we convert dates to int with a +, before converting
> toString(), we get 8x performance gain in Chrome and 20x performance gain
> in Firefox.
> In practice this means one can have tens of thousands of events in the
> calendar without noticing any responsiveness loss (and FullCalendar is
> awesome because it only processes events in the current view's date range).
> This only unknown variable is how Internet Explorer deals with it (I will
> check IE9 soon), as usual :)
> 
> Again thanks for the help on this as I believe that you are spot on.
> I'm using angular-ui in a commercial app (i.e. to make money), thanks for
> building this in the first place!!
> 
>  —
> Reply to this email directly or view it on GitHubhttps://github.com/angular-ui/angular-ui/pull/515#issuecomment-15642092
> .

## 

_Josh Kurz_

joshkurz:
@glebm thanks for this PR. Its really quite nice and a wonderful addition to the calendar. I think that it will most certainly make its way into the repo. 

 Could you re-open another PR here -> https://github.com/angular-ui/ui-calendar?

PR #519, opened 2013-03-28
'ui-validate-watch' causes field to be marked dirty immediately

rintaun:
The fix to #472, while otherwise awesome, causes the marked field to immediately be marked as dirty. This makes styling which relies on `ng-pristine` impossible.

### DIFF
--- a/modules/directives/validate/validate.js
+++ b/modules/directives/validate/validate.js
@@ -46,8 +46,10 @@ angular.module('ui.directives').directive('uiValidate', function () {
 
       // Support for ui-validate-watch
       if (attrs.uiValidateWatch) {
-        scope.$watch(attrs.uiValidateWatch, function() {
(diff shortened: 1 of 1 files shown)


ProLoser:
For shits and giggles, in the `$watch` try adding a `newVal, oldVal` parameters and do `if (newVal !== oldVal) ...`. This may prevent making the field dirty prematurely. Can you confirm?

rintaun:
Confirmed. I'll modify the file and convert this to a pull request.

rintaun:
It seems as though the unit test will need to be updated for this as well... I guess if I wrote tests as normally as I should, that would've occurred to me.

ProLoser:
Do you mind relocating this to https://github.com/angular-ui/ui-utils ?

ProLoser:
This project has been discontinued. You should reopen the issue if it's still relevant on angular-ui/ui-utils

Issue #469, opened 2013-03-02
uiCalendar > Re-creating the calendar on every update?


saurabhnanda:
If I understand https://github.com/angular-ui/angular-ui/blob/master/modules/directives/calendar/calendar.js#L37 correctly, is the directive basically destroying and re-creating the calendar every time `update()` is called? And `update()` is called whenever _anything_ changes in the events data?

Isn't this extremely inefficient? I understand if this is still WIP, and I'd be happy to contribute in making this better.

joshkurz:
Well pretty much that is what it is doing to keep angular and the calendar
in sync with each other. It seems that this is somewhat of the norm for
directives of this nature. The elm.html('') method is called every time
update is called, options are recreated, and then the fullcalendar method
is called again to re-sync the whole process.

yes this is still certainly a WIP, so if you have any suggestions or ideas
then please send some PR's so we can mull over them.

On Sat, Mar 2, 2013 at 3:08 AM, Saurabh Nanda notifications@github.comwrote:

> If I understand
> https://github.com/angular-ui/angular-ui/blob/master/modules/directives/calendar/calendar.js#L37correctly, is the directive basically destroying and re-creating the
> calendar every time update() is called? And update() is called whenever *
> anything\* changes in the events data?
> 
> Isn't this extremely inefficient? I understand if this is still WIP, and
> I'd be happy to contribute in making this better.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/angular-ui/angular-ui/issues/469
> .

## 

_Josh Kurz_
http://dillingermediaonline.com/wordpresshttp://dillingermediaonline.com

saurabhnanda:
@joshkurz : I'm unable to run `grunt` in the `angular-ui` source directory. I get the following error:

> grunt-cli: The grunt command line interface. (v0.1.6)
> Fatal error: Unable to find local grunt.
> If you're seeing this message, either a Gruntfile wasn't found or grunt
> hasn't been installed locally to your project. For more information about
> installing and configuring grunt, please see the Getting Started guide:
> http://gruntjs.com/getting-started

I installed node.js globally, and followed it up with the commands given below:

```
sudo npm install
sudo npm install -g grunt-cli
sudo npm install -g grunt
```

Any help would be appreciated.

ghinda:
@saurabhnanda You'll have to install the 0.3.x version of Grunt, locally.

Inside the angular-ui folder (where grunt.js is located):

```
[...]

PR #520, opened 2013-03-29
Reimplemented the event watch callback to remove and readd event instead...

Raibaz:
... of rebuilding the whole calendar.

This should fix issue #469 without breaking any of the previously implemented behaviours.

The existing tests seem to pass, and I think they will suffice without the need to create some specific ones for this new implementation.


joshkurz:
This looks Good to me. I would like to play with it some to make sure everything is cool. 

Good job. Thanks.

joshkurz:
Thanks for this PR. 

Im going to close this for now however, because we are not going to be adding any more additions to the calendar from the main angularUI repo. Could you re-open another PR here -> https://github.com/angular-ui/ui-calendar?

Raibaz:
Sure, done :)

PR #540, opened 2013-04-19
Fix initialization for select2 in AJAX mode

jmou:
Without this patch initialization gets broken in weird ways. The element value is set to the string "[object Object]" and initSelection is invoked to initialize the model value. If initSelection is not set as an option, the model value is set to null.


ProLoser:
AngularUI is currently undergoing a restructure. This project will no longer hold the codebase, can you please relocate this to https://github.com/angular-ui/ui-select2 ?

PR #546, opened 2013-04-23
Added true to the $watch expression on calendar directive

eibrahim:
This will allow the equalsTracker to detect changes even if the length of the arrays don't change.


eibrahim:
wrong fix... sorry.

glebm:
@eibrahim The current version of UI Calendar has fully automatic binding (without equalsTracker) and is available here https://github.com/angular-ui/ui-calendar

PR #554, opened 2013-04-29
Fix issue with uiDateFormat handling of epoch timestamps

jrschumacher:
Condition to check if value `isNumber` and dateFormat is `@`


ProLoser:
AngularUI is currently undergoing a project restructure.

Can you relocate this to https://github.com/angular-ui/ui-date (after confirming it's still necessary)?

jrschumacher:
Will do! angular-ui/ui-date#7

@ProLoser I would suggest adding this comment to the CONTRIBUTING.md to ease the burden on you ;)

PR #556, opened 2013-05-02
Allow user to control data binding in select2

rwlogel:
Adding an binder method to the select2 config which will allow caller to decided what value should be set in the view model using the 'data' returned from select2.


ProLoser:
AngularUI is currently undergoing a restructure. This project will no longer hold the codebase, can you please relocate this to https://github.com/angular-ui/ui-select2 ?

